Issue #74, opened 2015-03-17
add default permissions while saving a model


Bartzi:
For now permissions that have been assigned to a model object will not be assigned to it, when a new is created. This needs to be changed.

PR #76, opened 2015-03-24
allow automatically adding permissions to objects for groups

Bartzi:
contributes to #74 
still needs some tests that i will add later, today =)


_1327/documents/__init__.py
```diff
@@ -0,0 +1,2 @@
+# don't ask me why we need that... without it django will run into an error
```
janno42: i will definitely not accept such comments in the code. please either look up why we need it or at least write "needed to avoid error ..."

janno42:
after changing the comment: :+1:

PR #80, opened 2015-03-24
add tests to the default permission code

Bartzi:
entirely fixes #76 by adding tests and putting the signal intialisation code into the right spot as descibed in https://docs.djangoproject.com/en/1.7/topics/signals/


invliD:
:+1:

janno42:
:+1:

PR #81, opened 2015-03-24
Fixed reversion with polymorphic models

invliD:
See http://django-reversion.readthedocs.org/en/latest/api.html#multi-table-inheritance


janno42:
:+1:

PR #87, opened 2015-03-24
MenuItems should exclusively reference a Document or a Link

kaifabian:
MenuItemAdminForm now prevents setting link and document reference at the same time


Bartzi:
:+1:

PR #95, opened 2015-03-24
Improved document diff design

### LINKED ISSUE #48: Radiobutton list for version selection
For selecting two versions in the diff tool a list of all versions with two columns of radio buttons for selection would be nice.

Roughly like this: 
capture

invliD:
Closes #48.


Bartzi:
you could add some tests ;)
if you don't want to add a new ticket for tests...
but all in all: :+1:

PR #99, opened 2015-03-29
Added coffee-script installation to puppet manifest

### LINKED ISSUE #98: Add coffeescript to Vagrant
make sure that coffeescript gets automatically installed, wenn running vagrant up for the first time

invliD:
Fixes #98.

### DIFF
--- a/deployment/site.pp
+++ b/deployment/site.pp
@@ -26,7 +26,7 @@ node default {
     } ->
     exec { "install less":
         provider => shell,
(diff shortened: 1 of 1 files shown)


janno42:
from my experience that's not enough for ubuntu. i also had to install it via apt-get.
@Bartzi what do you think?

Bartzi:
I've tested it.. it works =) :+1:

PR #107, opened 2015-04-13
add function to create new info pages

### LINKED ISSUE #34: Autosave for protocols
The protocol editor should auto save content in case our beloved computer crashes.

Nef10:
fix #34 you can restore autosaves of new sites

### DIFF
--- a/_1327/information_pages/views.py
+++ b/_1327/information_pages/views.py
@@ -1,12 +1,13 @@
 from django.shortcuts import render
 from django.core.urlresolvers import reverse
 from django.contrib import messages
-from django.http import HttpResponseRedirect, HttpResponse
+from django.http import HttpResponseRedirect, HttpResponseForbidden, HttpResponse
 from django.utils.translation import ugettext_lazy as _
+from django.utils.text import slugify
 from django.forms.formsets import formset_factory
-from guardian.decorators import permission_required_or_403
 from guardian.shortcuts import get_perms
 from guardian.models import Group
+from datetime import datetime
 
 from _1327.documents.utils import handle_edit, prepare_versions, handle_autosave
 from _1327.documents.models import Document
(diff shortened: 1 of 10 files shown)


invliD:
Please rebase once #109 has been merged.

_1327/information_pages/views.py
```diff
@@ -14,6 +16,11 @@
 from _1327.information_pages.models import InformationDocument
 from _1327.user_management.shortcuts import get_object_or_error
 
+def create(request):
+	title = "New Page from "+str(datetime.now());
```
invliD: Make this translatable.

_1327/main/context_processors.py
```diff
@@ -31,3 +31,6 @@ def mark_selected(request, menu_item):
 	if current_view == item_view:
 		menu_item.selected = True
 		return True
+
+def can_create_informationpage(request):
+	return {'CAN_CREATE_INFORMATIONPAGE': request.user.has_perm("information_pages.add_informationdocument")}
```
invliD: Modify this to a more general "Create document" variant, see below.

_1327/templates/base.html
```diff
@@ -48,6 +48,13 @@
 						{% endif %}
 					</li>
 					{% endfor %}
+					{% if CAN_CREATE_INFORMATIONPAGE %}
+						<li class="pull-right">
+							<a href="{% url 'information_pages:create' %}">
+								<span class="glyphicon glyphicon-plus" aria-hidden="true"></span>
+							</a>
+						</li>
+					{% endif %}
```
invliD: We can't create such a link for each type of document, see above.
What about one link if the user can create *any* type of document?

janno42: i would say it's correct to have this here for info pages only. the only other types planned for now are transcripts and polls. both of them should be somehow collected in a list on specific pages (a list of transcripts and a list of polls) and the option to add a new page of this type can then be added on these pages.

invliD: Agreed.

invliD:
:+1:

Issue #104, opened 2015-04-13
permissions in templates


Bartzi:
When opening an information document with a non admin user, but with the correct permissions to edit the information page the according buttons are not there

PR #113, opened 2015-04-15
File uploading

### LINKED ISSUE #36: Allow adding attachments to a document
It should be possible to upload attachments for documents such as information pages or protocols.
An attachment could be:
* images
* pdfs
* excel sheets
* word documents
* ...

Bartzi:
Fixes #36 
It is now possible to add attachments to documents, attachments can be uploaded and will be saved in a media folder, attachments can be deleted and downloaded. This PR also includes tests for that functionality.

**Please Note**
The download feature needs the server to use the mod_sendfile extension as the file download first needs to be checked for sufficient permissions and is then handled by the webserver (this is a nice way i think)

### DIFF
--- a/_1327/documents/templates/attachments.html
+++ b/_1327/documents/templates/attachments.html
@@ -1,0 +1,101 @@
+{% load i18n %}
+{% load filename %}
+
+{% block information_page_content %}
+	<h2>{% trans "Add an attachment" %}</h2>
+
+	{% if form.errors %}
+		<div class="alert alert-danger alert-margin-bottom" role="alert">{% trans "Please provide the missing information" %}</div>
+	{% endif %}
+
+	<form class="form-horizontal" action="{{ edit_url }}" method="post" role="form" enctype="multipart/form-data">
+		{% csrf_token %}
+		<div class="form-group {% if 'displayname' in form.errors %}has-error{% endif %}">
+			<label for="inputFile" class="col-sm-2 control-label">{% trans "Displayname" %}</label>
+			<div class="col-sm-9">
+				<input type="text" class="form-control" name="displayname" value="{{ form.data.displayname }}" placeholder="{% trans "Optional displayname..." %}">
+			</div>
+		</div>
+		<div class="form-group {% if 'file' in form.errors %}has-error{% endif %}">
+			<label for="inputFile" class="col-sm-2 control-label">{% trans "Attachment" %}</label>
+			<div class="col-sm-9">
+				<input type="file" class="form-control" name="file" value="{{ form.data.file }}" placeholder="{% trans "Your Attachment..." %}">
+			</div>
+		</div>
+		<div class="form-group">
+			<div class="col-sm-offset-2 col-sm-4">
+				<button type="submit" class="btn btn-primary">
+					<i class="glyphicon glyphicon-floppy-disk"></i> {% trans 'Save' %}
+				</button>
+			</div>
+		</div>
+	</form>
+
+	<h2>{% trans "Already uploaded attachments" %}</h2>
+	<div class="col-sm-12">
+		<table class="table table-striped">
+			<thead>
+				<th>{% trans "Displayname" %}</th>
+				<th>{% trans "Filename" %}</th>
+				<th>{% trans "Created" %}</th>
+			</thead>
+			{% for attachment in attachments %}
+				<tr>
+					<td>
+						<input type="hidden" class="attachmentinput" name="file" value="{{ attachment.id }}">
+						<a href="{% url 'documents:download_attachment' %}?attachment_id={{ attachment.id }}">{{ attachment.displayname }}</a>
+					</td>
+					<td>{{ attachment.file.name|filename }}</td>
+					<td>{{ attachment.created }}</td>
+					<td>
+						<button type="button" data-displayname="{{ attachment.displayname }}" data-toggle="modal"
+						        data-target="#confirmAttachmentDeleteModal" class="delete-attachment btn btn-warning">
+							{% trans "Delete" %}
+						</button>
+					</td>
+				</tr>
+
+			{% endfor %}
+		</table>
+	</div>
+
+	<div class="modal fade" id="confirmAttachmentDeleteModal" tabindex="-1" role="dialog" aria-labelledby="myModalLabel" aria-hidden="true">
+		<div class="modal-dialog">
+			<div class="modal-content">
+				<div class="modal-header">
+					<button type="button" class="close" data-dismiss="modal" aria-label="Close">
+						<span aria-hidden="true">&times;</span>
+					</button>
+					<div class="modal-title">{% trans "Please Confirm" %}</div>
+				</div>
+				<div class="modal-body">
+					{% blocktrans %}
+					Do you really want to delete the Attachment <span id="attachment-name"></span>?
+					{% endblocktrans %}
+				</div>
+				<div class="modal-footer">
+					<button type="button" class="btn btn-primary" id="accept-delete">{% trans "Yes" %}</button>
+					<button type="button" class="btn btn-default" data-dismiss="modal">{% trans "No" %}</button>
+				</div>
+			</div>
+		</div>
+	</div>
+
+	<script>
+		$('#confirmAttachmentDeleteModal').on('show.bs.modal', function (event){
+			var button = $(event.relatedTarget);
+			var input = button.closest('tr').find('input.attachmentinput');
+			$('#attachment-name').html(button.data['displayname']);
+            var deleteButton = $('#accept-delete');
+            deleteButton.off('click');
+			deleteButton.on('click', function (event){
+				$.post("{% url "documents:delete_attachment" %}", {'id': input.attr('value')}, function (data, textStatus, jqXHR) {
+					location.reload()
+				});
+			});
+		});
+	</script>
+
+
+{% endblock %}
+
--- a/_1327/documents/tests.py
+++ b/_1327/documents/tests.py
@@ -1,4 +1,6 @@
+import tempfile
 from django.contrib.auth.models import Group
+from django.core.files.base import ContentFile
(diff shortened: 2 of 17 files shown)


janno42:
Please make sure that the tests delete the uploaded files even if an error occurs during execution.

_1327/documents/templates/attachments.html
```diff
@@ -0,0 +1,62 @@
...
+						<button type="button" class="delete-attachment btn btn-warning">{% trans "Delete" %}</button>
+					</td>
+				</tr>
+
+			{% endfor %}
+		</table>
+	</div>
+
+	<script>
+		$('.delete-attachment').on('click', function (event) {
```
janno42: deleting needs to have a confirmation

_1327/documents/tests.py
```diff
@@ -201,3 +203,187 @@ def test_view_permissions(self):
 
 			permission_names = [permission[0] for permission in subclass._meta.permissions]
 			self.assertIn(subclass.VIEW_PERMISSION_NAME, permission_names, msg="All Document subclasses must declare the permission named in VIEW_PERMISSION_NAME")
+
+
+class TestAttachments(WebTest):
+	"""
+		Tests that test the work with attachment.
+		Informationdocuments are used as testclass. It is assumed that the behaviour is similar with other documenttypes
```
janno42: Please use correct capitalization for classes, "behavior", and change the first sentence to "Tests creating, viewing and deleting attachments".

_1327/documents/tests.py
```diff
@@ -201,3 +203,187 @@ def test_view_permissions(self):
...
+
+	def tearDown(self):
+		self.attachment.file.delete()
+
+	def test_create_attachment(self):
+		upload_files = [
+			('file', 'test.txt', bytes(tempfile.SpooledTemporaryFile(max_size=10000, prefix='txt', mode='r')))
+		]
+
+		# test that user that has no change permission on a document can not add an attachment
```
janno42: user **who**, can not ... **neither** ...

_1327/documents/tests.py
```diff
@@ -201,3 +203,187 @@ def test_view_permissions(self):
...
+		self.assertEqual(response.status_code, 403)
+
+		response = self.app.post(reverse('information_pages:attachments', args=[self.document.url_title]),
+									content_type='multipart/form-data',
+									upload_files=upload_files,
+									user=normal_user,
+									expect_errors=True)
+		self.assertEqual(response.status_code, 403)
+
+		# test that user that is allowed to view the document may not add attachments to it
```
janno42: see above

_1327/documents/tests.py
```diff
@@ -201,3 +203,187 @@ def test_view_permissions(self):
 
 			permission_names = [permission[0] for permission in subclass._meta.permissions]
 			self.assertIn(subclass.VIEW_PERMISSION_NAME, permission_names, msg="All Document subclasses must declare the permission named in VIEW_PERMISSION_NAME")
+
+
+class TestAttachments(WebTest):
```
janno42: check whether all of the empty lines all over the file are necessary

_1327/documents/tests.py
```diff
@@ -201,3 +203,187 @@ def test_view_permissions(self):
...
+		self.assertEqual(response.status_code, 403)
+
+		response = self.app.post(reverse('information_pages:attachments', args=[self.document.url_title]),
+									content_type='multipart/form-data',
+									upload_files=upload_files,
+									user=normal_user,
+									expect_errors=True)
+		self.assertEqual(response.status_code, 403)
+
+		# test that member of group that has according permissions is allowed to upload attachments
```
janno42: member of group **who**

_1327/documents/tests.py
```diff
@@ -201,3 +203,187 @@ def test_view_permissions(self):
...
+									user=self.group_user)
+		self.assertEqual(response.status_code, 200)
+
+		response = self.app.post(reverse('information_pages:attachments', args=[self.document.url_title]),
+									content_type='multipart/form-data',
+									upload_files=upload_files,
+									user=self.group_user)
+		self.assertEqual(response.status_code, 200)
+
+		# test that superuser is allowed to upload attachments and see the corresponding page
```
janno42: **to** see

_1327/documents/tests.py
```diff
@@ -201,3 +203,187 @@ def test_view_permissions(self):
...
+
+		response = self.app.get(reverse('documents:delete_attachment'), params=params, expect_errors=True, xhr=True)
+		self.assertEqual(response.status_code, 404, msg="GET Requests are not allowed to work")
+
+		response = self.app.post(reverse('documents:delete_attachment'), params=params, expect_errors=True)
+		self.assertEqual(response.status_code, 404, msg="Requests that are not AJAX should return a 404 error")
+
+		response = self.app.post(reverse('documents:delete_attachment'), params=params, expect_errors=True, xhr=True)
+		self.assertEqual(response.status_code, 403,
+						msg="If a user has no permissions he should not be able to delete an attachment")
```
janno42: strange indentation?

_1327/documents/tests.py
janno42: please replace all "if the user ... he ..." with "if users ... they ..."

_1327/documents/tests.py
```diff
@@ -201,3 +203,187 @@ def test_view_permissions(self):
...
+		self.group.save()
+
+		self.group_user = UserProfile.objects.create_user("groupuser", "test", "test@test.test")
+		self.group_user.groups.add(self.group)
+		self.group_user.save()
+
+		self.document = InformationDocument(author=self.user, title="test", text="blabla")
+		self.document.save()
+
+		attachment_file = ContentFile("test content of test attachment")
```
janno42: could you check against the content below?

_1327/information_pages/templates/information_pages_base.html
```diff
@@ -17,6 +17,7 @@
 			<div class="divider"></div>
 			<a class="btn btn-sm btn-sidebar {% if active_page == 'view' %}btn-warning{% else %}btn-default{% endif %}" href="{% url 'information_pages:view_information' document.url_title %}" role="tab">{% trans "View" %}</a>
 			<a class="btn btn-sm btn-sidebar {% if active_page == 'edit' %}btn-warning{% else %}btn-default{% endif %}" href="{% url 'information_pages:edit' document.url_title %}" role="tab">{% trans "Edit" %}</a>
+			<a class="btn btn-sm btn-sidebar {% if active_page == 'attachments' %}btn-warning{% else %}btn-default{% endif %}" href="{% url 'information_pages:attachments' document.url_title %}" role="tab">{% trans "Attachments" %}</a>
```
janno42: here only staff users can see the button

janno42:
deleting attachments does not remove the file from the server. is this intended?

janno42:
the attachment link on the attachments page links to ```/information/attachments/test/documents/00/00/FILENAME/``` which doesn't exist

janno42:
downloading a test attachment returned an empty file (0kb) :(

janno42:
The order of attachments must be configurable

janno42:
The filename of the downloaded file should not have an added ```documents-00-00-``` in the beginning

_1327/documents/templates/attachments.html
```diff
@@ -0,0 +1,62 @@
...
+				<button type="submit" class="btn btn-primary">
+					<i class="glyphicon glyphicon-floppy-disk"></i> {% trans 'Save' %}
+				</button>
+			</div>
+		</div>
+	</form>
+
+	<h2>{% trans "Already uploaded attachments" %}</h2>
+	<div class="col-sm-offset-2 col-sm-9">
+		<table class="table table-striped">
```
janno42: this table should include the filename and the date when the file was uploaded

_1327/information_pages/templates/information_pages_base.html
Bartzi: sure this is another problem that already has a ticket ;) see #104

Bartzi:
downloading an attachment using the django development server returns a file with 0KB... this is due to the fact that a real webserver should deliver the requested static file after django checked that the user has the permissions he needs. For this I've chosen to use sendfile, which is a mod to Apache and allows doing this. First step is to check whether the user has sufficient permissions to download the requested document and then it sends a response with the "x-sendfile" header set that tells the webserver to deliver the requested file to the user.
The django development server is not capable of understanding this and thus delivers an empty file.

invliD:
I don't like the use of sendfile. And I didn't actually find it. Do you mean mod_xsendfile?

Bartzi:
Yes I do mean mod_xsendfile ;) But we can also change it... I just implemented it that way because i thought this is the most django way... serving the 'static' file via a webserver but still checking for permissions using django. But I'm open for another option =)

Bartzi:
okay I think I've addressed all comments... deleting an attachment now requires the user to confirm that, the download link is fixed, the files are removed from the server, tests will remove all attachments even if they fail, I've changed the comments and I changed the downloading thing... 
I've added a new library 'django-sendfile' (https://github.com/johnsensible/django-sendfile) that helps us handling the whole thing in a nice and easy way which is able to work with different webservers and still works with the original idea ;) I hope you guys like it, too =)

**If you want to try it make sure to update your pip stuff using the requirements.txt file**

janno42:
The filename on the attachment page should be ```file.name``` instead of ```documents/DD/MM/file.name```

janno42:
If the displayname has no file ending it should get the one from the original file

janno42:
Uploading a file now throws an error for me:
```
'NoneType' object has no attribute 'group'
/vagrant/_1327/information_pages/views.py in attachments
	success, form = handle_attachment(request, document)
/vagrant/_1327/documents/utils.py in handle_attachment
	file_type = re.search(r'\.(\w{3})$', instance.file.name).group(1) 
```

janno42:
:+1:

PR #122, opened 2015-05-04
fix permissions in templates

### LINKED ISSUE #104: permissions in templates

Bartzi:
fixes #104


invliD:
:+1:

PR #131, opened 2015-05-04
Added reversion support for ForeignKeyFields and ManyToMany fields

invliD:


Bartzi:
Looks good so far... How do you handle ForeignKeys? I can only see ManyToManyFields...

invliD:
In the following line:
``if hasattr(document_class, key):``
Only ForeignKeys and ManyToManyFields exist as attributes of the class. So anything that's not ManyToMany must be a ForeignKey.

Bartzi:
alright :smile: then :shipit:

PR #133, opened 2015-05-05
adjust style of markdown editor

### LINKED ISSUE #132: Markdown editor style
The markdown editor produces a few weird artifacts:
bildschirmfoto 2015-05-05 um 16 22 31
The label "Text" has a red font and on the top and bottom of the textarea there's a ``border: 1px dashed red``.

Nef10:
fix #132 
@invliD please have a look

### DIFF
--- a/_1327/static/less/1327.less
+++ b/_1327/static/less/1327.less
@@ -243,3 +243,24 @@ table.table-narrow {
 .toc-heading {
 	font-weight: bold;
 }
+
+.md-editor {
+	border-radius: @input-border-radius;
+}
+
+.has-error .md-editor {
+	border-color: @state-danger-text;
+	.box-shadow(inset 0 1px 1px rgba(0,0,0,.075));
+}
+
+.has-error .md-editor.active {
+	border-color: darken(@state-danger-text, 10%);
+	@shadow: inset 0 1px 1px rgba(0,0,0,.075), 0 0 6px lighten(@state-danger-text, 20%);
+	.box-shadow(@shadow);
+}
+
+.md-editor textarea {
+	padding: @padding-base-vertical @padding-base-horizontal;
+	border-bottom: none;
+	border-top-style: solid;
+}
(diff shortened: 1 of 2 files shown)


_1327/static/less/1327.less
```diff
@@ -226,3 +226,17 @@ footer a {
...
+	box-shadow: inset 0 1px 1px rgba(0, 0, 0, 0.075);
+}
+
+.md-editor {
+	border-radius: 4px;
+}
+
+.md-editor textarea {
+	padding: 6px 12px
+}
```
invliD: Please add a newline.

_1327/static/less/1327.less
```diff
@@ -226,3 +226,17 @@ footer a {
...
 }
+
+.has-error .md-editor {
+	border-color: #b1063a;
+	-webkit-box-shadow: inset 0 1px 1px rgba(0, 0, 0, 0.075);
+	box-shadow: inset 0 1px 1px rgba(0, 0, 0, 0.075);
+}
+
+.md-editor {
+	border-radius: 4px;
```
invliD: Use bootstrap variables for all values.

invliD:
The entire editor still has a red border and "Text" is also still red. Why does the form control have the bootstrap class ``has-error`` when creating a new document?
The textarea also still has a dashed border on top and on the bottom, it's just gray now.

Nef10:
Because it is a mandatory field, I did not changed this.
The dashed border is the style of the editor (see http://www.codingdrama.com/bootstrap-markdown/)

invliD:
But it's already red when you open the page, not just after submitting without text.
I think you should remove the dashed border. It looks kinda weird.

Nef10:
The problem is caused by the form validation and not by my styles.

Nef10:
But another problem: we want to style the div with the editor and not textarea itself on focus. Unfortunately this requires javascript because the css selector is not supported yet: http://css4-selectors.com/selector/css4/subject-of-selector-with-child-combinator/

_1327/static/less/1327.less
```diff
@@ -226,3 +226,17 @@ footer a {
 table.table-narrow {
 	max-width: 400px;
 }
+
+.has-error .md-editor {
+  .form-control-validation(@state-danger-text; @state-danger-text; @state-danger-bg);
```
invliD: What does this do? I can't tell the difference between using this and commenting it out.

Nef10:
Regarding error when creating a new document see #148.
Finished now

invliD:
Please rebase on top of master instead of merging master to your feature branch.

PR #134, opened 2015-05-06
Minutes

invliD:
ref #2 

@Bartzi Do you have an idea on how to fix the ``test_group_permission_hook`` test? It tries to instantiate and then save all subclasses of ``Document``, but it's not that easy for the new ``MinutesDocument`` since it has some required fields.


Bartzi:
Actually I don't really know... I thought about checking every field whether it is required or not and according to that change the data supplied to the create call. But this is actually no viable option because there are also a lot of other fields in a model that we should not change by messing around with the meta object... 
We could add this test for every subclass of its own... but that is really ugly... 

The actual problem is that the foreign keys are not mapped that is why the database complaints with a `NOT NULL` constraint violation... maybe we can find a way to trick the system into thinking that they are provided correctly....

karyon:
https://github.com/rbarrois/factory_boy and https://github.com/vandersonmota/model_mommy are projects that generate fixtures at runtime. you can specify attributes where you need a specific value, and they randomly make up all other attributes based on the field type. if you don't need any specific values, they should also be able to generate models completely randomly, that might be useful here.

btw that's something i wished we started tests in EvaP with instead of those jsons.

invliD:
Sounds cool, I'll look into it.

Bartzi:
yeah might be the solution to the problem :smile: quite cool stuff

invliD:
I tried to use factory_boy, but it turns out you still need to manually define ForeignKeys. It only auto-detects regular fields. model_mommy seems to have the same problem.

karyon:
did you have a look at [the second example](https://model-mommy.readthedocs.org/en/latest/basic_usage.html#model-relationships) in the model_mommy docs? it uses foreign keys and it looks like it should work.

invliD:
Seems like you're right. I'll try that soon.

invliD:
That was easy, thanks! It's done now.

invliD:
Do not merge, I have to add what #141 and #146 did for information pages.

invliD:
This should do it.

_1327/minutes/templates/minutes_base.html
```diff
@@ -0,0 +1,71 @@
...
+{% block sidebar %}
+	<dl class="meta-information">
+		<dt>{% trans "Moderator" %}</dt>
+		<dd>{{ document.moderator }}</dd>
+		<dt>{% trans "Minute taker" %}</dt>
+		<dd>{{ document.author }}</dd>
+		<dt>{% trans "Participants" %}</dt>
+		<dd>
+			{% for participant in document.participants.all %}
+			<div>{{ participant }}</div>
```
Nef10: just use one dd per participant

_1327/minutes/templates/minutes_attachments.html
```diff
@@ -0,0 +1,9 @@
+{% extends 'minutes_base.html' %}
+
+{% block title %}
```
Bartzi: I guess we don't need that block as it is alredy declared like that in the base template `minutes_base.html`

_1327/minutes/templates/minutes_edit.html
```diff
@@ -0,0 +1,9 @@
+{% extends 'minutes_base.html' %}
+
+{% block title %}
```
Bartzi: same thing here as in `minutes_attachments.html`

_1327/minutes/templates/minutes_permissions.html
```diff
@@ -0,0 +1,9 @@
+{% extends 'minutes_base.html' %}
+
+{% block title %}
```
Bartzi: same thing here as in `minutes_attachments.html`

_1327/minutes/templates/minutes_versions.html
```diff
@@ -0,0 +1,15 @@
+{% extends 'minutes_base.html' %}
+{% load static %}
+
+{% block title %}
```
Bartzi: same thing here as in `minutes_attachments.html`

requirements.txt
```diff
@@ -13,3 +13,4 @@ django-polymorphic >= 0.6.1
 django-sendfile >= 0.3.6
 
 Markdown >= 2.6.0
+model_mommy
```
Bartzi: could you add a min. version? Just that it looks like the other lines...

Bartzi:
when you are done with all comments you are free to go :smile:  but do not forget to work on #156

PR #139, opened 2015-05-08
Added regular design to Django admin

invliD:
All html templates were copied from ``django-admin-bootstrapped`` and then modified to fix issues or to better suit our needs.

### DIFF
--- a/_1327/documents/admin.py
+++ b/_1327/documents/admin.py
@@ -14,7 +14,7 @@ class DocumentAdmin(GuardedModelAdmin, reversion.VersionAdmin, PolymorphicParent
 		(InformationDocument, InformationDocumentAdmin),
 	)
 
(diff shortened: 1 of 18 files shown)


Bartzi:
Nice job! But there are some things that look a little bit odd for me... maybe it is just me but well lets see:

The text on the dark part is hardly readable:
image

The striped table on the overview page should have different colours when there is more than one item:

striped_admin

Could you have a look at this and fix it?

invliD:
I also noticed both of these issues. I didn't fix them because they are just bugs in our bootstrap theme, I talked to @janno42 about fixing that afterwards. I'd create an issue once this PR has been merged.

Bartzi:
okay well then create that issue and merge this thing ;)

PR #140, opened 2015-05-08
Format all the things

invliD:
I used this command to find all the issues:
``find _1327 -name migrations -prune -a -name static -prune -o -name "*.py" -exec pep8 --ignore=W191,E501 {} \;``


Bartzi:
Looks good to me =) :+1: :shipit:

PR #141, opened 2015-05-09
Added document TOC to information page sidebar

### LINKED ISSUE #129: TOC in sidebar
In view mode the sidebar of information pages should have a TOC of the main headlines of the page linking to the respective position on the page.

invliD:
Page markdown is now rendered in Python instead of JavaScript.

Closes #129.

@janno42: You'll have to make it look nice. ;)

### DIFF
--- a/_1327/information_pages/templates/information_pages_base.html
+++ b/_1327/information_pages/templates/information_pages_base.html
@@ -10,6 +10,7 @@
 {% endblock %}
 
 {% block sidebar %}
+	{{ toc|safe }}
 	<dl class="meta-information">
 		<dt>{% trans "Author" %}</dt>
 		<dd>{{ document.author }}</dd>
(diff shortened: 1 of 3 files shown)


Bartzi:
I'm not sure whether using the safe filter is a good idea because it opens the door for XSS attacks...
Have a look at the following example:

image

This may be saved and everything and when looking at the page the following happens:

image

Maybe there is another way to make that markdown work in that way (which is actually a nice way)?

invliD:
Huh, you're right. But is this an actual problem? Users can never create content. Only admins and staff can.
We could simply strip out (or escape) HTML before converting Markdown to HTML, that would also fix it. We have to use the ``safe`` filter here since there's the converted HTML in the ``text`` variable.

Bartzi:
I think although users should never be able to create any content we should make sure that this is not possible. Therefore I vote for escaping the HTML before converting to Markdown.

invliD:
It's done. Turns out python-markdown has support for that. I'll just do a quick rebase.

Bartzi:
awesome :smile: go for it :+1:

PR #145, opened 2015-05-10
color changes

### LINKED ISSUE #143: Fix striped tables and panel footers
For screenshots see #139.

janno42:
fix #143; change colors of striped tables, panel footers and primary buttons

### DIFF
--- a/_1327/static/less/1327.less
+++ b/_1327/static/less/1327.less
@@ -8,6 +8,9 @@
 .modal-dialog {
 	z-index: @zindex-modal-background + 1;
 }
+.table-striped > tbody > tr:nth-of-type(even) {
+	background-color: @table-bg-accent-even;
+}
 // End fix.
 
 // Fix jsdifflib line wrapping:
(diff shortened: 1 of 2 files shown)


_1327/static/less/1327.less
```diff
@@ -220,3 +220,11 @@ footer a {
 table.table-narrow {
 	max-width: 400px;
 }
+
+// Overwrite bootstrap table coloring
+.table-striped > tbody > tr:nth-of-type(odd) {
+	background-color: @white;
+}
```
invliD: Please modify ``@table-bg-accent`` and remove this part.

_1327/static/less/1327.less
```diff
@@ -220,3 +220,11 @@ footer a {
...
+.table-striped > tbody > tr:nth-of-type(even) {
+	background-color: @gray-lighter;
+}
```
invliD: Please introduce a new variable for this color (right next to ``@table-bg-accent``), put in a comment that it's not a bootstrap variable, and move this block to the bootstrap fixes at the top of the file.

_1327/static/less/1327.less
janno42: i still need the definition for even rows. i thought having both at the same place would be nice. what do you think?

invliD: See my other comment below.

janno42:
updated

invliD:
:+1:

PR #149, opened 2015-05-10
Fixed attachment deletion

invliD:
Important part is the second change. It fixes an XSS vulnerability and the deletion itself. ``data`` needs to be accessed using round brackets.


Nef10:
:+1:

PR #166, opened 2015-05-18
add margin to headings in sidebar

### LINKED ISSUE #165: add padding to headings in sidebar
so that e.g. author and created are more separated

Nef10:
fix #165

### DIFF
--- a/_1327/static/less/1327.less
+++ b/_1327/static/less/1327.less
@@ -240,6 +240,10 @@ table.table-narrow {
 	margin-bottom: 0;
 }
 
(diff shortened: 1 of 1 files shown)


_1327/static/less/1327.less
```diff
@@ -240,6 +240,10 @@ table.table-narrow {
 	margin-bottom: 0;
 }
 
+.sidebar dt:not(:first-child) {
+  padding-top: 8px;
```
invliD: Use margin instead.

janno42:
:+1:

PR #168, opened 2015-05-18
Fixed document create/edit form to not validate unless the user tries to save

### LINKED ISSUE #148: do not validate empty form when creating a new site
otherwise the mandatory fields will be marked red

invliD:
Fixes #148.

### DIFF
--- a/_1327/documents/utils.py
+++ b/_1327/documents/utils.py
@@ -5,7 +5,7 @@ from django.db import transaction
 import reversion
 
 from _1327.documents.models import TemporaryDocumentText
-from _1327.documents.forms import TextForm, AttachmentForm
+from _1327.documents.forms import DocumentForm, AttachmentForm
 
 def get_new_autosaved_pages_for_user(user):
 	autosaved_pages = []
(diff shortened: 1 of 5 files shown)


_1327/documents/utils.py
```diff
@@ -41,26 +39,24 @@ def handle_edit(request, document):
...
 			autosaved = False
 
-		form_data = {
-			'title': document.title,
-			'text': text,
-		}
-		form = TextForm(form_data)
+		if 'restore' in request.GET and autosave is not None:
+			form_data = {
+				'text': autosave.text,
```
Nef10: you loose the old title of the page if you restore an autosave

invliD: It's now fixed in line 57. (The use of the ``initial`` dictionary overwrites the text, all other fields are still taken from the ``document`` model)

_1327/information_pages/models.py
```diff
@@ -25,3 +26,10 @@ def get_edit_url(self):
 		return reverse('information_pages:edit', args=(self.url_title, ))
 
 reversion.register(InformationDocument, follow=["document_ptr"])
+
+
+class InformationDocumentForm(DocumentForm):
```
Bartzi: is there a reason why this is in the `models.py`?

Bartzi:
:+1:

PR #170, opened 2015-05-18
Added a list of minutes

invliD:


_1327/minutes/views.py
```diff
@@ -16,6 +17,24 @@
 from .models import MinutesDocument
 
 
+@login_required
+def list(request):
+	years = {}
+	minutes = MinutesDocument.objects.order_by('-date')
+	for m in minutes:
```
Bartzi: could you name it `minute` instead of `m`?

invliD: You don't use the singular. Minutes are always plural.

Bartzi: oh okay in that case...

Bartzi:
:+1:

PR #186, opened 2015-07-13
Fixed navigation context processor in case a view does not exist

### LINKED ISSUE #182: 404s produce 500s instead if DEBUG is False
Need to fix main.context_processors.mark_selected.

invliD:
Fixed #182.


Bartzi:
:+1:

PR #195, opened 2015-07-13
fix #180 add information for non public pages

### LINKED ISSUE #180: Message on non-public pages
Information pages should show a message at some point if they are not visible for anonymous users.
This is to prevent people from creating pages and forgetting to give the Anonymous group the right to see them.

Nef10:


_1327/information_pages/views.py
```diff
@@ -76,12 +78,18 @@ def view_information(request, title):
 	md = markdown.Markdown(safe_mode='escape', extensions=[TocExtension(baselevel=2)])
 	text = md.convert(document.text)
 
+	anonymous = get_user_model().objects.get(id=settings.ANONYMOUS_USER_ID)
```
Bartzi: please use `get_anonymous_user()` from `guardian.utilities`

Bartzi:
after you've adjusted the tooltip: :+1:

janno42:
:+1:

PR #211, opened 2015-10-18
Fixed static precompiler

invliD:
We didn't pin the version of django-static-precompiler. A few of the template tags and config changed, this PR fixes the issues.


janno42:
:+1:

PR #247, opened 2016-02-01
show site in german

Nef10:


_1327/templates/master.html
```diff
@@ -27,3 +30,4 @@
 {% endblock %}
 </body>
 </html>
+{% endlanguage %}
```
Bartzi: please add a newline ;)

janno42:
:de:

PR #214, opened 2015-11-09
add guests to minutes

Nef10:
fix #206 

@Bartzi This copies the ```MinutesDocumentForm``` into ```minutes/models.py```, because importing ```minutes/forms.py``` is not possible due circular references. Could you have a look at this?

PR #249, opened 2016-02-02
Fixed Django 1.9 deprecation warnings

invliD:
I removed the import that you added in #214, @Nef10. I can't remember what the problem was, but I couldn't find anything that breaks after removing it now.


Nef10:
Currently it works after removing.
But if we are going to remove `GuestForm` from `forms.py` in the future, this file will no longer be included in the project, and the `MinutesDocumentForm` will be back to standard.

Nef10:
fixed in #289

PR #250, opened 2016-02-02
Moved sidebar to the top on mobile

invliD:


Bartzi:
works but the page as such looks quite terrible on a mobile device...
:+1:

PR #266, opened 2016-02-15
rename flake8 command to lint

### LINKED ISSUE #264: rename flake8 command to lint
Makes more sense to name it lint. Does not imply which linter is used.

Bartzi:
fixes #264 

makes it clear that we are linting.


invliD:
:+1:

PR #283, opened 2016-02-16
Fix all the things

janno42:
see commit messages for changes

### DIFF
--- a/_1327/main/context_processors.py
+++ b/_1327/main/context_processors.py
@@ -2,7 +2,7 @@ from . import models
 
 
 def menu(request):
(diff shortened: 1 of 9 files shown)


_1327/main/context_processors.py
```diff
@@ -1,8 +1,17 @@
+from django.utils import translation
+
 from . import models
 
 
+def set_language(request):
+	user_language = 'de'  # always set to German
```
Nef10: could we put this string into the settings?

janno42: done

Nef10:
:+1:

PR #285, opened 2016-02-17
fix #267; allow third navigation level

### LINKED ISSUE #267: Third navigation level
The navigation needs a third level.

janno42:
add third navigation level


Nef10:
Could we show an error if the admin tries to add a 4th menu level oder a 2nd in the footer?
Otherwise :+1:

janno42:
now only first and second level items can be selected as parent and there is a check that footer items don't have a parent.

Nef10:
:+1:

PR #291, opened 2016-02-22
Permissions for Polls

Bartzi:
This PR enables us to provide permissions for polls, thereby refactoring the Permission form to be more general and easy to adapt to new situations.

The code has two pitfalls:

1. I was not able to come up with a nicer rendering of permission names for each group. Does s.b. have a better idea?
2. I had to replace the `as_table ` method of the form. This is because I did not see another way of rendering the form in a generic way and the standard as_table method does not render the html the way we want. If somebody has any suggestions...

Please also make sure to review whether I checked that all permissions are checked at the right place at the right time.

I will add tests later on. That's why this PR is WIP

### DIFF
--- a/_1327/documents/forms.py
+++ b/_1327/documents/forms.py
@@ -3,6 +3,7 @@ from django.conf import settings
 from django.contrib.auth.models import Group, Permission
 from django.core.exceptions import ValidationError
 from django.forms import BooleanField
+from django.utils.encoding import force_text
 from django.utils.safestring import mark_safe
 from django.utils.translation import ugettext_lazy as _
 from guardian.shortcuts import assign_perm, get_perms, remove_perm
@@ -61,10 +62,11 @@ class PermissionBaseForm(forms.BaseForm):
 	def header(cls):
 		output = [
 			'<tr>',
-			'<th class="col-md-6"> {} </th>'.format(_("Role")),
+			'<th class="col-md-6">{}</th>'.format(_("Role")),
 		]
+		# add one table head entry for every permission in this form
 		for permission in sorted([name for name, field in filter(lambda x: type(x[1]) == BooleanField, cls.base_fields.items())]):
-			item = "<th class=\"col-md-2 text-center\"> {} </th>".format(_(permission.rsplit('_')[0]))
+			item = "<th class=\"col-md-2 text-center\">{}</th>".format(force_text(_(permission.rsplit('_')[0])))
 			output.append(item)
 		output.append('</tr>')
 		return mark_safe('\n'.join(output))
@@ -73,14 +75,14 @@ class PermissionBaseForm(forms.BaseForm):
 		"Returns this form rendered as HTML <tr>"
 		output = [
 			"<tr>",
-			"<td>{}</td>".format(self['group_name'].value())
+			"<td>{}</td>".format(force_text(self['group_name'].value()))
 		]
 
 		for name in sorted(self.fields.keys()):
 			if name == "group_name":
 				continue
-			output.append('<td class="text-center"> {} </td>'.format(self[name]))
-		output.append(str(self['group_name']))
+			output.append('<td class="text-center">{}</td>'.format(force_text(self[name])))
+		output.append(force_text(str(self['group_name'])))
 		output.append("</tr>")
 
 		return mark_safe('\n'.join(output))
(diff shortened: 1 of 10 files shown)


_1327/polls/templates/polls_create_poll.html
```diff
@@ -39,6 +39,19 @@
...
+        <h3>{% trans "Permissions" %}</h3>
+        <div class="divider"></div>
+        {{ permission_formset.management_form }}
+        <table class="table table-striped vertically-aligned table-narrow" >
+			<thead>
+				{{ permission_header }}
+			</thead>
+			<tbody>
+			{% for form_element in permission_formset %}
+                {{ form_element.as_table }}
```
Nef10: use tabs

_1327/documents/forms.py
```diff
@@ -40,39 +41,70 @@ def clean_url_title(self):
...
+			else:
+				remove_perm(field_name, group, model)
+
+	@classmethod
+	def header(cls, content_type):
+		output = [
+			'<tr>',
+			'<th class="col-md-6"> {} </th>'.format(_("Role")),
+		]
+		for permission in sorted(Permission.objects.filter(content_type=content_type), key=lambda x: x.codename):
```
invliD: Try using the fields instead of a separate database query.

_1327/documents/forms.py
```diff
@@ -40,39 +42,70 @@ def clean_url_title(self):
...
+			if value:
+				assign_perm(field_name, group, model)
+			else:
+				remove_perm(field_name, group, model)
+
+	@classmethod
+	def header(cls):
+		output = [
+			'<tr>',
+			'<th class="col-md-6"> {} </th>'.format(_("Role")),
```
invliD: Remove spaces around `{}`

_1327/documents/forms.py
```diff
@@ -40,39 +42,70 @@ def clean_url_title(self):
...
+		]
+		for permission in sorted([name for name, field in filter(lambda x: type(x[1]) == BooleanField, cls.base_fields.items())]):
+			item = "<th class=\"col-md-2 text-center\"> {} </th>".format(_(permission.rsplit('_')[0]))
```
invliD: Same here.

_1327/documents/forms.py
```diff
@@ -40,39 +42,70 @@ def clean_url_title(self):
...
+			else:
+				remove_perm(field_name, group, model)
+
+	@classmethod
+	def header(cls):
+		output = [
+			'<tr>',
+			'<th class="col-md-6"> {} </th>'.format(_("Role")),
+		]
+		for permission in sorted([name for name, field in filter(lambda x: type(x[1]) == BooleanField, cls.base_fields.items())]):
```
invliD: Though I could figure out what this does, please add a comment.

_1327/documents/forms.py
```diff
@@ -40,39 +42,70 @@ def clean_url_title(self):
...
+			item = "<th class=\"col-md-2 text-center\"> {} </th>".format(_(permission.rsplit('_')[0]))
+			output.append(item)
+		output.append('</tr>')
+		return mark_safe('\n'.join(output))
```
invliD: If you mark this as safe, you should escape all the stuff going in the HTML.

_1327/documents/forms.py
```diff
@@ -40,39 +42,70 @@ def clean_url_title(self):
...
+		]
+
+		for name in sorted(self.fields.keys()):
+			if name == "group_name":
+				continue
+			output.append('<td class="text-center"> {} </td>'.format(self[name]))
+		output.append(str(self['group_name']))
+		output.append("</tr>")
+
+		return mark_safe('\n'.join(output))
```
invliD: Same here.

_1327/documents/forms.py
```diff
@@ -40,39 +42,70 @@ def clean_url_title(self):
...
+		"Returns this form rendered as HTML <tr>"
+		output = [
+			"<tr>",
+			"<td>{}</td>".format(self['group_name'].value())
+		]
+
+		for name in sorted(self.fields.keys()):
+			if name == "group_name":
+				continue
+			output.append('<td class="text-center"> {} </td>'.format(self[name]))
```
invliD: Same here.

janno42:
the `add` permission should not be shown in the table

janno42:
when any permission for a group gets selected, the clean method should check that the `view` permission is also selected.

janno42:
the `permission_warning` (documents/utils.py) is defined for InformationDocuments, so when used for minutes the warning is always shown.

PR #295, opened 2016-02-29
Minutes publish button

Nef10:
refs #2 and refs #298
Should this create a new revision?


janno42:
seems good :+1:
i don't think we need a new revision for permission changes.

PR #301, opened 2016-03-07
enable editing menu items outside of the admin area

janno42:


_1327/main/templates/menu_item_edit.html
```diff
@@ -0,0 +1,63 @@
...
+	<script type="text/javascript">
+		$(document).ready(function() {
+			$("#id_document,#id_parent,#id_menu_type").select2({
+				language: "{{ LANGUAGE_CODE }}",
+				placeholder: "{% trans "Please select..." %}",
+				minimumResultsForSearch: 3
+			});
+		});
+	</script>
+{% endblock %}
```
Bartzi: newline missing ;)

_1327/main/views.py
```diff
@@ -8,6 +10,9 @@
 from _1327.documents.models import Document
 from _1327.documents.utils import permission_warning
 
+from .forms import MenuItemForm
```
Bartzi: do we want relative imports here? Normally we are using absolute imports... we can change this, but then we should change it everywhere

invliD: We actually do this in a lot of places if both modules are in the same Django app, so I think it's ok.

Bartzi: really? But if you say so this seems to be okay ;)

Bartzi:
:+1:

_1327/main/views.py
invliD:
```bash
…/1327/_1327:master* % grep -ri "from \." .
./documents/admin.py:from .models import Attachment, Document
./documents/forms.py:from .models import Attachment, Document
./documents/tests.py:from .models import Attachment, Document
./information_pages/forms.py:from .models import InformationDocument
./main/admin.py:from . import models
./main/context_processors.py:from . import models
./main/tests.py:from .context_processors import mark_selected
./main/tests.py:from .models import MenuItem
^C
```

PR #332, opened 2016-06-06
cache import statements of get_model_function

### LINKED ISSUE #328: Cache get_model_function calls
We should come up with a caching strategy that avoids completing the import of a python module whenever views like `documents:edit` are called. This might slow down the system as a whole. 

We might create a dictionary that maps the name of  an already imported function to the function itself.

This feature has been intrdocued in #326

Bartzi:
fixes #328


janno42:
:zap:

PR #333, opened 2016-06-06
Anonymous user must not get other permissions than view

### LINKED ISSUE #306: Anonymous user must only get view permissions
currently anonymous users can vote for polls or edit documents if they have the permission set, but django runs into an error because no real user exists in the request.

anonymous users should not be allowed to vote or edit anyway, so it must be made sure that only the view permission can be given to the anonymous user's group.

Nef10:
fix #306


janno42:
:rabbit:

PR #334, opened 2016-06-06
show correct groups when editing documents

### LINKED ISSUE #327: False value in groups field on minutes document editing
When editing a minutes document which belongs to multiple groups, the groups field in the form only has one group selected.

Nef10:
fix #327


janno42:
:palm_tree:

PR #337, opened 2016-06-11
Django 1.9

janno42:
update to django 1.9

### DIFF
--- a/_1327/urls.py
+++ b/_1327/urls.py
@@ -1,21 +1,24 @@
-from django.conf.urls import include, patterns, url
+from django.conf.urls import include, url
 from django.contrib import admin
+
+from _1327.main import views as main_views
+from _1327.user_management import views as user_management_views
 
 admin.autodiscover()
 
-urlpatterns = patterns('_1327',  # noqa
-	url(r"^$", 'main.views.index', name='index'),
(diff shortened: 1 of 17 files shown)


_1327/urls.py
```diff
@@ -1,21 +1,24 @@
...
+from _1327.user_management import views as user_management_views
+
 admin.autodiscover()
 
-urlpatterns = patterns('_1327',  # noqa
-	url(r"^$", 'main.views.index', name='index'),
+urlpatterns = [
+	url(r"^$", main_views.index, name='index'),
 	url(r"^documents/", include('_1327.documents.urls', namespace='documents')),
 	url(r"^minutes/", include('_1327.minutes.urls', namespace='minutes')),
```
invliD: What about also importing these and referencing the actual urls module instead of a string here?

janno42: i think this is the preferred way, at least according to the docs (https://docs.djangoproject.com/en/1.9/topics/http/urls/#including-other-urlconfs)

invliD: I don't see any comment about which form is preferred, but it's fine either way, I guess.

invliD:
:+1:

PR #339, opened 2016-06-13
make choices of polls sortable

### LINKED ISSUE #191: make choices of polls sortable
Choices of polls should be sortable. See #188 for a possible implementation. Parent of this issue is #175

Bartzi:
make choices of polls sortable
fixes #191

### DIFF
--- a/_1327/polls/templates/polls_edit.html
+++ b/_1327/polls/templates/polls_edit.html
@@ -4,74 +4,74 @@
 {% load bootstrap3 %}
 
 {% block inlineformset %}
-    {{ formset.management_form }}
-    <h3>Choices</h3>
-    {% bootstrap_formset_errors formset layout='inline' %}
-    <div id="choice-inline-form">
-        {% for form_element in formset %}
-            <div class="choice-row">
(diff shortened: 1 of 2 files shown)


_1327/polls/templates/polls_edit.html
```diff
@@ -4,74 +4,74 @@
...
+	});
+
+	// make forms sortable
+	$('#choice-inline-form table tbody').sortable({
+		items: '.sortable',
+		handle: '.sorthandle',
+	});
+
+	// add datepicker to date inputs
+	$("input[name*='date']:not([readonly='True'])").datepicker( $.datepicker.regional["{{ LANGUAGE_CODE }}"] );
```
janno42: just copied although not needed?

janno42:
nice. evap helps ;)
see the one comment, then :+1:

_1327/polls/templates/polls_edit.html
Bartzi: its actually necessary right now. We would not have a datepicker without this line ;) but this is something we could put into to master template for editing documents

PR #342, opened 2016-06-13
restore autosaves for each document type

### LINKED ISSUE #336: Restore must be per document type
When e.g. creating a new minutes document and having an unsaved poll, one currently gets the option to restore that poll, although it's a different document type.

Bartzi:
fixes #336 

autosave restore suggestion is now only for the corresponding document type


janno42:
:+1:

invliD:
:+1:

PR #343, opened 2016-06-13
allow users with only view permission to see result of polls

### LINKED ISSUE #185: Enable view_poll without participate_in_poll
It should be possible to allow users seeing a poll but not participating in it.
Thus an additional permission should be implemented.

Bartzi:
fixes #185

### DIFF
--- a/_1327/polls/views.py
+++ b/_1327/polls/views.py
@@ -42,9 +42,7 @@ def list(request):
 	)
 
 
-def results(request, url_title):
-	poll = get_object_or_error(Poll, request.user, ['polls.view_poll'], url_title=url_title)
-
+def results(request, poll, url_title):
 	if poll.start_date > datetime.date.today():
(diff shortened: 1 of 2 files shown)


_1327/polls/views.py
```diff
@@ -42,17 +42,15 @@ def list(request):
...
+def results(request, poll, url_title):
 	if poll.start_date > datetime.date.today():
 		# poll is not open
 		raise Http404
 
 	if request.user.has_perm(Poll.VOTE_PERMISSION_NAME, poll) \
 				and not poll.participants.filter(id=request.user.pk).exists() \
 				and poll.end_date > datetime.date.today():
-		return vote(request, url_title)
+		return vote(request, url_title, poll)
```
janno42: please keep the same argument order

janno42:
after the one thing :+1:

PR #349, opened 2016-06-27
fix default url_title generation for polls

### LINKED ISSUE #346: Default url title for polls is not unique
When creating polls, the url title is always `polls_new-poll`. This fails when creating a poll, saving it without url changes and creating another one.
```
MultipleObjectsReturned at /documents/poll/create
get() returned more than one Document -- it returned 2!
```

the url title should become a unique field and the default value should include the timestamp like infopages.

Bartzi:
fixes #346


janno42:
unique needed in model

_1327/polls/models.py
```diff
@@ -53,7 +53,13 @@ def generate_new_title(cls):
 
 	@classmethod
 	def generate_default_slug(cls, title):
-		return "polls_{}".format(slugify(title))
+		slug_start = "polls_{}".format(slugify(title))
```
janno42: please use `new-poll-from-` and add the date instead of `polls_` and adding the title

janno42: `slug_final = slug = "new-poll-from-{}".format(date.today().isoformat())`

janno42: i would suggest the form from https://github.com/fsr-itse/1327/pull/351

janno42:
:+1:

PR #352, opened 2016-06-27
save votes for each choice atomically

### LINKED ISSUE #348: Save poll votes atomically
Currently, the number of votes is read, incremented and then saved without using transactions. The Django ORM allows using field references using `F`. We should use that to increment the vote counter in one query.

Bartzi:
fix #348


janno42:
:+1:

invliD:
:+1:

PR #353, opened 2016-06-27
filter users while querying for temporary document texts

### LINKED ISSUE #347: Temporary documents are filtered in Python
Currently, the autosave check retrieves all temporary documents and then filters in python. At least the author check should be moved to the DB.

Bartzi:
fix #347


janno42:
:+1:

PR #354, opened 2016-06-27
put datepicker initialisation into documents base edit template

### LINKED ISSUE #350: put initialisation of datepickers into base template
right now we have two different document types that need datepickers. The initialisation of the datepickers is done in two different locations. We should do this in one space (maybe in the base template of the page)

Bartzi:
fix #350


janno42:
:+1:

PR #355, opened 2016-07-11
use python markdown rendering instead of javascript library

### LINKED ISSUE #293: change preview to also use python markdown parser
We cannot guarantee that our two parser return the same result

Bartzi:
fixes #293


Nef10:
👍

PR #365, opened 2016-08-06
make system work with flake 8 >= 3.0

### LINKED ISSUE #362: Adapt Code to work with flake8 3.0.0
The new version of flake 8 seems to work differently than before. The linting command is not working when using flake8 3.0.0. We should fix this.

Bartzi:
fix #362


janno42:
:+1:

PR #367, opened 2016-08-06
Make it possible to choose whether results of a poll shall be shown immediately after voting or not

### LINKED ISSUE #345: Results of Polls should

Bartzi:
fixes #345


_1327/polls/views.py
```diff
@@ -53,6 +53,13 @@ def results(request, poll, url_title):
 				and poll.end_date > datetime.date.today():
 		return vote(request, poll, url_title)
 
+	if not poll.show_results_immediately and poll.end_date > datetime.date.today():
```
janno42: please use consistently `poll.end_date >= datetime.date.today()` to check whether the voting period is still in progress. users should be able to vote until and including the `end_date`

janno42:
please remove the link for polls in the results lists where the results can not be accessed and add an info sign with tooltip and put the explaining message there.

PR #369, opened 2016-08-08
fix function to create menu item

### LINKED ISSUE #358: Create menu items
- [ ] show the "Manage menu items" entry in the user menu for all users who can edit menu items
- [ ] allow access to the page only to these users
- [ ] add option to define parent element
- [ ] allow only parent elements that are on level 0 or level 1 and for which the user has the `change_children` permission and make sure that non-superusers do always specify a parent

- [ ] ~~add option to define permissions for newly created menu items (e.g., select a group that will get permissions like for minutes)~~ (not needed because parent element defines that the user can edit this item)

Nef10:
fix #358

### DIFF
--- a/_1327/main/context_processors.py
+++ b/_1327/main/context_processors.py
@@ -1,6 +1,8 @@
 from django.conf import settings
 from django.utils import translation
+from guardian.shortcuts import get_objects_for_user
 
+from _1327.main.models import MenuItem
 from . import models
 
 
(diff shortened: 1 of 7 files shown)


Bartzi:
looks good so far, but I would like to see some tests ;)

Bartzi:
as I said, I'd like it more with tests, after you've opened a ticket for the tests you can :shipit:

janno42:
for tests see #373

PR #380, opened 2016-08-15
Anonymous user authentication

### LINKED ISSUE #15: Authentication
Users need to be able to login using both Kerberos and additional (plain Django) users.

invliD:
This PR adds a mapping from IP range to group ID that allows anonymous users to gain the permissions of the group if they are in the correct IP range.

Closes #15.

### DIFF
--- a/_1327/settings.py
+++ b/_1327/settings.py
@@ -35,6 +35,12 @@ ANONYMOUS_GROUP_NAME = "Anonymous"
 STAFF_GROUP_NAME = "Staff"
 UNIVERSITY_GROUP_NAME = "University Network"
 
+# Anonymous users in one of the given IP ranges are automatically assumed to be in the associated group.
+# Only the first match will be used.
+ANONYMOUS_IP_RANGE_GROUPS = {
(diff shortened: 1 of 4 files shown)


Bartzi:
:ship:

_1327/settings.py
```diff
@@ -73,12 +74,19 @@
...
 ]
 
 # needed by django-guardian library
 ANONYMOUS_USER_ID = -1
 GUARDIAN_RAISE_403 = True
 
+# Anonymous users in one of the given IP ranges are automatically assumed to be in the associated group.
+# Only the first match will be used.
+ANONYMOUS_IP_RANGE_GROUPS = {
+	# Example: '127.0.0.0/8': 2,
```
janno42: could you do this by group name? we have defined `UNIVERSITY_GROUP_NAME` in the settings. and if we need a custom group to be added it's still easier to recognize by name.

invliD:
@janno42: Done.

janno42:
:+1:

PR #382, opened 2016-08-15
allow impersonation of users

### LINKED ISSUE #366: View as functionality
We need to have a view as functionality for admin users to easily check permissions of other users/user groups for menu items, documents, etc.
A possible solution might be https://github.com/arteria/django-hijack

Bartzi:
You want to work as a different user and do stuff you always wanted to do?
Now is your Chance!! (Only for superusers...)

Fixes #366

### DIFF
--- a/_1327/templates/master.html
+++ b/_1327/templates/master.html
@@ -1,6 +1,7 @@
 {% load static %}
 {% load compile_static %}
 {% load bootstrap3 %}
+{% load hijack_tags %}
 
 <!DOCTYPE html>
 <html>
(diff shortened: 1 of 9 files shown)


janno42:
:+1:

_1327/templates/master.html
```diff
@@ -13,6 +13,7 @@
 		<link rel="shortcut icon" type="image/x-icon" href="{{STATIC_URL}}/images/favicon.ico"/>
 		<link rel="stylesheet" href="//ajax.googleapis.com/ajax/libs/jqueryui/1.11.4/themes/smoothness/jquery-ui.css">
 		<link rel="stylesheet" href="https://cdnjs.cloudflare.com/ajax/libs/select2/4.0.2/css/select2.min.css" />
+		<link rel="stylesheet" type="text/css" href="{% static 'hijack/hijack-styles.css' %}" />
```
Nef10: load only if user is hijacked

PR #396, opened 2016-09-24
fix and modify document creation

### LINKED ISSUE #388: Document creation for non-superusers
Currently users who are not superusers can not create documents (minutes, information pages, polls) if they only have the create permission for these objects, because the creation redirects to the edit view that checks the edit permission. General edit permissions per object type should not be granted, so we need a different solution here.
Creating a document could give an additional parameter to the edit view stating that no edit permission is needed. However, once saving the document, the user won't be able to access it again, so the user can't even grant own permissions. The last-mentioned is an issue for all documents besides minutes, because these are assigned to a group on creation. Perhaps we should assign initial permissions on creation also for the other document types?

Nef10:
fix #388


_1327/documents/forms.py
```diff
@@ -44,6 +56,12 @@ def clean_url_title(self):
 			raise ValidationError(_('The URL used for this page is already taken.'))
 		return url_title
 
+	def clean_group(self):
+		value = self.cleaned_data['group']
+		if value and value not in self.user_groups:
+			raise ValidationError(_("You are not a member in this group!"))
```
Bartzi: "You are not a member of this group?"

_1327/minutes/views.py
```diff
@@ -1,41 +1,24 @@
...
-	groups = {}
-	minutes = MinutesDocument.objects.order_by('-date')
+def list(request, groupid):
+	groupid = int(groupid)
+	try:
+		group = Group.objects.get(id=groupid)
+	except ObjectDoesNotExist:
+		raise SuspiciousOperation
+	result = {}
+	minutes = get_objects_for_group(group, "minutes.change_minutes", MinutesDocument).order_by('-date')
```
Bartzi: permission name should be "minutes.change_minutesdocument"

_1327/minutes/views.py
```diff
@@ -1,41 +1,24 @@
...
-def list(request, groupid=None):
-	if groupid:
-		groupid = int(groupid)
-		try:
-			Group.objects.get(id=groupid)
-		except ObjectDoesNotExist:
-			raise SuspiciousOperation
-	groups = {}
-	minutes = MinutesDocument.objects.order_by('-date')
+def list(request, groupid):
```
Bartzi: please explain what this function does and why we filter for the change permission

_1327/minutes/views.py
```diff
@@ -1,41 +1,24 @@
...
+	groupid = int(groupid)
+	try:
+		group = Group.objects.get(id=groupid)
+	except ObjectDoesNotExist:
+		raise SuspiciousOperation
+	result = {}
+	minutes = get_objects_for_group(group, "minutes.change_minutes", MinutesDocument).order_by('-date')
 	for m in minutes:
-		if not request.user.has_perm(MinutesDocument.get_view_permission(), m):
```
Bartzi: please add these lines again

PR #407, opened 2016-09-26
allow setting vote right while creation a poll

### LINKED ISSUE #395: allow selection of user with vote rights during poll creation

Nef10:
fix #395

### DIFF
--- a/_1327/documents/utils.py
+++ b/_1327/documents/utils.py
@@ -11,7 +11,6 @@ from reversion import revisions
 
 from _1327.documents.forms import AttachmentForm, DocumentForm
 from _1327.documents.models import Document, TemporaryDocumentText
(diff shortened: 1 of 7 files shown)


_1327/documents/utils.py
```diff
@@ -61,6 +63,12 @@ def handle_edit(request, document, formset=None, initial=None):
 			if not document.has_perms():
 				document.set_all_permissions(cleaned_data['group'])
 
+			if content_type == ContentType.objects.get_for_model(Poll):
```
Bartzi: how about adding a new function for each document that can be called to add additional perms while editing?
This should do nothing in the base class and can be overriden in each subclass

janno42: +1

PR #413, opened 2016-09-27
fix has perm check of document

Nef10:


_1327/documents/models.py
```diff
@@ -138,6 +138,8 @@ def has_perms(self):
 		group_perms = get_groups_with_perms(self, attach_perms=True)
 		for perms in group_perms.values():
 			for perm in perms:
+				content_type = ContentType.objects.get_for_model(self)
```
Bartzi: please put this right after line 138, so that it only gets evaluated once

Nef10:
@Bartzi changed

PR #418, opened 2016-09-27
allow selection of group with edit rights for menu items on creation

### LINKED ISSUE #412: select group with edit rights when creating a menu item

Nef10:
fix #412

### DIFF
--- a/_1327/main/forms.py
+++ b/_1327/main/forms.py
@@ -1,4 +1,6 @@
 from django import forms
+from django.conf import settings
+from django.contrib.auth.models import Group
 from django.contrib.contenttypes.models import ContentType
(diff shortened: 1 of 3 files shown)


_1327/main/forms.py
```diff
@@ -55,30 +57,62 @@ def clean(self):
 
 
 class MenuItemCreationForm(MenuItemForm):
+	group = forms.ModelChoiceField(Group.objects.all(), label=_('Edit permissions'), disabled=False, required=True)
```
Bartzi: we should think about putting all of this group handling stuff in the forms into a common place where it can also be used by the document forms

Nef10:
see #419

PR #428, opened 2016-10-29
add list of unlinked information pages

### LINKED ISSUE #427: List of information pages without menu entry
All users who can add information pages should get access to a page listing all the information pages for which the user has edit rights that are not accessible via a menu entry.
So new pages that have not been added to the menu can easily be spotted.

janno42:
fix #427


_1327/information_pages/views.py
```diff
@@ -0,0 +1,15 @@
+from django.shortcuts import render
+
+from guardian.shortcuts import get_objects_for_user
+
+from _1327.information_pages.models import InformationDocument
+
+
+# shows a list of information documents that are not accessible via a menu entry
+def unlinked_list(request):
+	permission_name = InformationDocument.objects.first().edit_permission_name  # the property can only be accessed from an instance of the class
```
Nef10: what happens if there is no information document?

janno42:
now there is a message if the list is empty

PR #432, opened 2016-11-03
don't show publish button on creation

janno42:


_1327/documents/views.py
```diff
@@ -88,7 +88,7 @@ def edit(request, title, new_autosaved_pages=None, initial=None):
 			'form': form,
 			'attachment_form': attachment_form,
 			'active_page': 'edit',
-			'creation': (len(revisions.get_for_object(document)) == 0),
```
Nef10: does this work after restoring an autosave

janno42: thanks, that made me checking it again and now using the same creation check everywhere

PR #442, opened 2016-11-14
change standard localization to english

Bartzi:
change standard localization during development to english

set localization to German in `localsettings.py` in production


Bartzi:
closing because obsoleted

PR #443, opened 2016-11-14
fix university network permissions

Bartzi:
fixes #421

PR #444, opened 2016-11-14
rework authentication backend to complete custom authentication backend

### LINKED ISSUE #421: University Network rights also for logged in users
Users who are logged in and who are connected from an IP in the University Network range should still get the permissions from the University Network Group. Currently these permissions are only granted for unauthenticated users.

Bartzi:
fixes #421 obsoletes #443

### DIFF
--- a/_1327/polls/models.py
+++ b/_1327/polls/models.py
@@ -57,7 +57,7 @@ class Poll(Document):
 	def get_vote_permission(klass):
 		content_type = ContentType.objects.get_for_model(klass)
 		app_label = content_type.app_label
-		permission = "{}.{}".format(app_label, klass.VOTE_PERMISSION_NAME)
+		permission = "{app}.{permission_name}".format(app=app_label, permission_name=klass.VOTE_PERMISSION_NAME)
 		return permission
 
 	@classmethod
@@ -71,8 +71,7 @@ class Poll(Document):
 
 	@property
 	def vote_permission_name(self):
-		content_type = ContentType.objects.get_for_model(self)
-		return "{app}.{permission_name}".format(app=content_type.app_label, permission_name=POLL_VOTE_PERMISSION_NAME)
+		return Poll.get_vote_permission()
 
 	def get_view_url(self):
 		return reverse('documents:view', args=(self.url_title,))
--- a/_1327/polls/models.py
+++ b/_1327/polls/models.py
@@ -54,6 +54,13 @@ class Poll(Document):
 		return _("New Poll")
 
 	@classmethod
+	def get_vote_permission(klass):
+		content_type = ContentType.objects.get_for_model(klass)
+		app_label = content_type.app_label
+		permission = "{}.{}".format(app_label, klass.VOTE_PERMISSION_NAME)
+		return permission
+
+	@classmethod
 	def generate_default_slug(cls, title):
 		slug_final = slug = "new-poll-from-{}".format(date.today().isoformat())
 		count = 2
--- a/_1327/user_management/authentication.py
+++ b/_1327/user_management/authentication.py
@@ -22,14 +22,14 @@ class _1327AuthorizationBackend:
 
 		group_name = user_obj._ip_range_group_name if hasattr(user_obj, '_ip_range_group_name') else None
 
-		has_perm = False
 		if user_obj.is_authenticated():
-			# user is not anonymous user and has no permissions on other elements
+			# user is not anonymous user and no other backend confirmed the permission yet
 			# --> we need to check the anonymous permissions again
 			check = ObjectPermissionChecker(get_anonymous_user())
-			has_perm = check.has_perm(perm, obj)
-		elif group_name:
+			if check.has_perm(perm, obj):
+				return True
+		if group_name:
 			group = Group.objects.get(name=group_name)
 			check = ObjectPermissionChecker(group)
-			has_perm = check.has_perm(perm, obj)
-		return has_perm
+			return check.has_perm(perm, obj)
+		return False
--- a/_1327/user_management/authentication.py
+++ b/_1327/user_management/authentication.py
@@ -1,9 +1,10 @@
 from django.contrib.auth.models import Group
 from django.contrib.contenttypes.models import ContentType
-from guardian.shortcuts import get_perms
+from guardian.core import ObjectPermissionChecker
+from guardian.utils import get_anonymous_user
 
 
-class IPRangeAuthorizationBackend:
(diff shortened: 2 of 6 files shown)


_1327/user_management/tests.py
```diff
@@ -102,3 +106,38 @@ def test_impersonate_wrong_url(self):
 
 		response = self.app.post('/hijack/{username}/'.format(username=user.username), user=self.user, expect_errors=True)
 		self.assertEqual(response.status_code, 400)
+
+
+class _1327AuthenticationBackendTests(WebTest):
```
Bartzi: Please have a Look at these tests again, as they do not really reflect the behaviour currently implemented in the authenticationbackend

_1327/polls/models.py
```diff
@@ -54,6 +54,13 @@ def generate_new_title(cls):
 		return _("New Poll")
 
 	@classmethod
+	def get_vote_permission(klass):
```
janno42: please refactor. there is `def vote_permission_name(self)` just two methods below.

_1327/user_management/authentication.py
```diff
@@ -1,29 +1,35 @@
...
+		# Only permissions from the same app as the object are valid.
+		content_type = ContentType.objects.get_for_model(obj)
+		if app != content_type.app_label:
+			return False
+
+		group_name = user_obj._ip_range_group_name if hasattr(user_obj, '_ip_range_group_name') else None
+
+		has_perm = False
+		if user_obj.is_authenticated():
+			# user is not anonymous user and has no permissions on other elements
```
janno42: `user is not anonymous user and no other backend confirmed the permission yet`

_1327/user_management/authentication.py
```diff
@@ -1,29 +1,35 @@
...
+			# --> we need to check the anonymous permissions again
+			check = ObjectPermissionChecker(get_anonymous_user())
+			has_perm = check.has_perm(perm, obj)
+		elif group_name:
```
janno42: why `elif`? shouldn't this test be done even if the user is authenticated?
please do an early return above if the user has the permission so that we can skip this part.

_1327/user_management/tests.py
```diff
@@ -102,3 +106,38 @@ def test_impersonate_wrong_url(self):
...
+
+	def setUp(self):
+		self.document = mommy.make(InformationDocument)
+		self.user = mommy.make(UserProfile)
+
+	def test_anonymous_fallback_if_user_has_no_permissions(self):
+		for group in Group.objects.all():
+			assign_perm(self.document.view_permission_name, group, self.document)
+
+		response = self.app.get(reverse('documents:view', args=[self.document.url_title]), expect_errors=True, user=self.user)
```
janno42: why `expect_errors=True`?

_1327/user_management/tests.py
```diff
@@ -102,3 +106,38 @@ def test_impersonate_wrong_url(self):
...
+
+	csrf_checks = False
+	extra_environ = {'HTTP_ACCEPT_LANGUAGE': 'en'}
+
+	def setUp(self):
+		self.document = mommy.make(InformationDocument)
+		self.user = mommy.make(UserProfile)
+
+	def test_anonymous_fallback_if_user_has_no_permissions(self):
+		for group in Group.objects.all():
```
janno42: shouldn't you just assign permissions for the anonymous user's group?

_1327/user_management/tests.py
```diff
@@ -102,3 +106,38 @@ def test_impersonate_wrong_url(self):
...
+			assign_perm(self.document.view_permission_name, group, self.document)
+
+		response = self.app.get(reverse('documents:view', args=[self.document.url_title]), expect_errors=True, user=self.user)
+		self.assertEqual(response.status_code, 200)
+
+	def test_anonymous_fallback_without_anonyomous_permission(self):
```
janno42: typo in second `anonymous`

_1327/user_management/tests.py
```diff
@@ -102,3 +106,38 @@ def test_impersonate_wrong_url(self):
...
+
+		response = self.app.get(reverse('documents:view', args=[self.document.url_title]), expect_errors=True, user=self.user)
+		self.assertEqual(response.status_code, 403)
+
+	def test_anonymous_fallback_not_used_if_user_has_permission(self):
+		group = mommy.make(Group)
+		self.user.groups.add(group)
+		self.user.save()
+
+		for group in Group.objects.all().exclude(name=settings.ANONYMOUS_GROUP_NAME):
```
janno42: please only grant permissions to `group`

_1327/user_management/tests.py
```diff
@@ -102,3 +106,38 @@ def test_impersonate_wrong_url(self):
...
+			assign_perm(self.document.view_permission_name, group, self.document)
+
+		response = self.app.get(reverse('documents:view', args=[self.document.url_title]), expect_errors=True, user=self.user)
```
janno42: again, why `expect_errors=True`?

_1327/user_management/tests.py
janno42: what exactly do you mean by "do not really reflect the behaviour currently implemented"?
and can you please add a comment somewhere that the university network can't be easily tested in here?

Bartzi: I just added this comment to remind myself that I still need to add tests for the university network. I just forgot them ;)

_1327/polls/models.py
Bartzi: yeah that's right, but that method below is no classmethod, and a classmethod was necessary, that's why I added this

_1327/user_management/authentication.py
Bartzi: good catch `elif` is totally wrong here, an early return is also a good idea

_1327/user_management/tests.py
Bartzi: yeah right ;)

_1327/user_management/authentication.py
```diff
@@ -1,29 +1,34 @@
...
+		if app != content_type.app_label:
+			return False
+
+		group_name = user_obj._ip_range_group_name if hasattr(user_obj, '_ip_range_group_name') else None
+
+		if user_obj.is_authenticated():
+			# user is not anonymous user and no other backend confirmed the permission yet
+			# --> we need to check the anonymous permissions again
+			check = ObjectPermissionChecker(get_anonymous_user())
+			return check.has_perm(perm, obj)
```
janno42: only return if true

PR #453, opened 2016-11-21
Add breadcrumbs

### LINKED ISSUE #449: Breadcrumbs
Documents should get breadcrumbs if they are accessible via the menu. This should contain links to the parent items from the menu.

First mockup:
capture

janno42:
fix #449 
adding breadcrumbs for documents that are accessible via the menu.

minutes, polls and information pages which are not accessible via the menu don't get breadcrumbs.


janno42:
done

PR #459, opened 2016-11-21
Multiselect, vote group selection, minutes author

### LINKED ISSUE #424: Allow changing minutes taker
It should be possible to change the minutes taker, e.g. when the person who took the minutes is not adding them to the system.

### LINKED ISSUE #436: Select2 for all multi selections
In all cases where multiple options can be selected from a list, a select2 widget should be used.

### LINKED ISSUE #457: Don't show vote group selection after poll creation
The vote group selection field should only be visible on a poll's creation. Afterwards the permission page can be used to make changes.
Currently it is always visible, but empty although permissions have already been set.

janno42:
fix #424: allow changing minutes taker
fix #436: use select2 for all multiselection fields
fix #457: show vote group field only on poll creation


janno42:
added tests

PR #470, opened 2016-12-18
Urls

janno42:
- allow slashes in custom urls
- prefix free urls for information pages
- use distinct url prefixes for minutes and polls


_1327/documents/models.py
```diff
@@ -45,6 +44,11 @@ def url(self, id):
 	def __str__(self):
 		return self.title
 
+	def save(self, *args, **kw):
```
Bartzi: please you rename `**kw` to `**kwargs`

PR #494, opened 2017-01-08
default user group

janno42:
add new users to default group


_1327/user_management/models.py
```diff
@@ -67,3 +69,11 @@ def __str__(self):
 	@cached_property
 	def is_staff(self):
 		return self.is_superuser or self.groups.filter(name=settings.STAFF_GROUP_NAME).exists()
+
+
+@receiver(post_save, sender=UserProfile)
+def add_to_default_group(sender, **kwargs):
+	if settings.DEFAULT_USER_GROUP_NAME and kwargs.get('created', False):
+		user = kwargs["instance"]
```
Bartzi: maybe you could write `user = kwargs.get("instance")` but that is just my personal taste when I work with kwargs, so it is not necessary ;)

PR #496, opened 2017-01-08
order participants by name

janno42:


_1327/main/templatetags/morefilters.py
```diff
@@ -24,3 +24,8 @@ def can_delete_menu_item(menu_item, user):
 @register.filter(name='can_view_menu_item')
 def can_view_menu_item(menu_item, user):
 	return menu_item.can_view(user)
+
+
+@register.filter(name='sort_users_by_name')
+def sort_users_by_name(users):
+	return sorted(users, key=lambda user: (user.last_name, user.first_name))
```
Bartzi: this fails if any user involved as a participant has no first or last name in the database. 
maybe we should include the username here, too?

Issue #429, opened 2016-11-03
Sort polls new to old


janno42:
Poll results should be sorted from new to old

janno42:
@PFischbeck do you already have a solution implemented? if not i'd like to take over.

PR #497, opened 2017-01-08
Sort polls by end date

PFischbeck:
This solves #429 and additionally reduces the number of SQL queries in the polls index list by a lot.

Is there a better/cleaner way to pass the attribute (whether the poll can be changed by the user) to the view?

### DIFF
--- a/_1327/polls/views.py
+++ b/_1327/polls/views.py
@@ -24,7 +24,7 @@ def index(request):
 	finished_polls = []
 	upcoming_polls = []
 	# do not show polls that a user is not allowed to see
-	for poll in Poll.objects.all():
+	for poll in Poll.objects.all().order_by('-end_date'):
 		if request.user.has_perm(Poll.get_view_permission(), obj=poll) and poll.start_date <= datetime.date.today():
 			if datetime.date.today() <= poll.end_date \
 						and not poll.participants.filter(id=request.user.pk).exists() \


_1327/polls/templates/polls_index.html
```diff
@@ -109,9 +107,8 @@
 						{% endif %}
 						<td>{{ poll.start_date }} - {{ poll.end_date }}</td>
 						<td class="text-right">
-							{% get_obj_perms request.user for poll as "poll_perms" %}
-							{% if "change_poll" in poll_perms %}
-								<a class="btn btn-warning btn-xs" href="{% url poll.get_edit_url_name poll.url_title %}">{% trans "Edit Poll" %}</a>
+							{% if poll.changeable %}
+                                <a class="btn btn-warning btn-xs" href="{% url poll.get_edit_url_name poll.url_title %}">{% trans "Edit Poll" %}</a>
```
janno42: please use tabs for indentation

_1327/polls/views.py
```diff
@@ -17,22 +17,32 @@
...
 
 def index(request):
 	running_polls = []
 	finished_polls = []
 	upcoming_polls = []
+
+	polls = Poll.objects.all().order_by('-end_date').prefetch_related('participants')
+
+	# Prefetch user permissions
+	user_checker = ObjectPermissionChecker(request.user)
```
janno42: the problem here is that you only check for permissions the user has, but we need to check the group permissions and ip range group permissions as well (see `minutes/views.py:20`).
before we change the permission checks in too many places (before we only had `has_perm` which is overwritten to fulfill our needs) we need to come up with a general solution. maybe you could discuss that with @Bartzi.

PFischbeck:
This now only sorts the polls, without any SQL query optimization. Should we open an issue for the poll list perf optimization?

janno42:
yes, please open a ticket and keep what you did so far.

PR #503, opened 2017-01-14
CSS fixes

### LINKED ISSUE #502: accessing some menu items is not possible ...
... when said menu is only accessible via an area that goes over the "logout" button.
Then the menu loses focus, closes and the "logout" menu pops up.

This can be resolved by giving the "logout" button a negative z-index.

janno42:
- change login button's z-index (fix #502)
- prevent flickering from login button size change on hovering, then moving into a new line and thus not being hovered anymore, changing back to old size and the old position
- remove whitespace from breadcrumbs block

### DIFF
--- a/_1327/documents/templates/documents_base.html
+++ b/_1327/documents/templates/documents_base.html
@@ -47,7 +47,7 @@
 	{% endif %}
 {% endblock %}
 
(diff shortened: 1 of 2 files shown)


janno42:
see #515

PR #513, opened 2017-02-06
make it possible to use urls with trailing slash or without

### LINKED ISSUE #506: Trailing slashes in URL
URLs should be resolvable with or without trailing slashes.

Bartzi:
fix #506


janno42:
I don't really like URLs with trailing slashes, so I would prefer using URLs without slash, it should only be possible to access the page even with a trailing slash.

Bartzi:
Yes we can do this...
I could implement it using [this](https://djangosnippets.org/snippets/2655/) middleware...

janno42:
and what about `.../?` in the regexes?

Bartzi:
very good idea! That actually seems to work quite well! Please have a look again

janno42:
it does not work with information pages (`.../infopage/` does not resolve if the url title is `infopage`)

Bartzi:
yeah right... well in this case I do not see any other possibilty, but implementing the middleware because we can not really add this to URLs we chose by ourselves...

janno42:
Well, we could adapt the document retrieval and remove trailing slashes in the view.

Bartzi:
obsoleted by #519

PR #515, opened 2017-02-06
Menu UI changes

janno42:
see commit messages

### DIFF
--- a/_1327/templates/base.html
+++ b/_1327/templates/base.html
@@ -54,14 +54,10 @@
 							</li>
 						{% endif %}
 					{% endfor %}
-					<li class="dropdown pull-right{% if user.is_authenticated %} active{% endif %} login-button">
-						<a href="{% if user.is_authenticated %}{% url 'logout' %}{% else %}{% url 'login' %}{% endif %}" class="login-link"><span class="glyphicon glyphicon-user"></span></a>
-						<ul class="dropdown-menu">
-							{% if user.is_authenticated %}
-								<li class="disabled"><a href="#">{% trans "Logged in as" %} {{ user.username }}</a></li>
-							{% endif %}
-							{% if CAN_CREATE_INFORMATIONPAGE or CAN_CREATE_MINUTES or CAN_CREATE_POLL %}
-								<li role="separator" class="divider"></li>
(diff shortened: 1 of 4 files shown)


_1327/user_management/tests.py
```diff
@@ -134,7 +134,7 @@ def test_impersonate_any_user(self):
 			response = response.follow()
 			self.assertEqual(response.status_code, 200)
 
-			self.assertIn("Logged in as {username}".format(username=user.username), response.body.decode('utf-8'))
+			self.assertIn("{username}".format(username=user.username), response.body.decode('utf-8'))
```
Bartzi: I guess we should change this `username=user.username` to  this `username=user.get_full_name()`

_1327/templates/base.html
```diff
@@ -54,14 +54,10 @@
...
 					{% endfor %}
-					<li class="dropdown pull-right{% if user.is_authenticated %} active{% endif %} login-button">
-						<a href="{% if user.is_authenticated %}{% url 'logout' %}{% else %}{% url 'login' %}{% endif %}" class="login-link"><span class="glyphicon glyphicon-user"></span></a>
-						<ul class="dropdown-menu">
-							{% if user.is_authenticated %}
-								<li class="disabled"><a href="#">{% trans "Logged in as" %} {{ user.username }}</a></li>
-							{% endif %}
-							{% if CAN_CREATE_INFORMATIONPAGE or CAN_CREATE_MINUTES or CAN_CREATE_POLL %}
-								<li role="separator" class="divider"></li>
+					{% if CAN_CREATE_INFORMATIONPAGE or CAN_CREATE_MINUTES or CAN_CREATE_POLL %}
```
Bartzi: If a user is not allowed to create information pages, minutes or polls, but allowed to  edit menu items he can not see the menu. Is this intended?

janno42:
admin area layout is nothing that needs to be fixed, so i wouldn't change anything there

_1327/templates/base.html
janno42: there is no use case where a user is allowed to edit menu items but not allowed to add information pages, that's why i thought it would be okay

_1327/user_management/tests.py
janno42: good catch. fixed.

PR #524, opened 2017-02-17
Render markdown tables as HTML table

### LINKED ISSUE #430: Support markdown tables
we need to support displaying markdown tables

Nef10:
fix #430


janno42:
note to self: css

PR #550, opened 2017-03-13
Handle multiple temporary documents

### LINKED ISSUE #549: Handle multiple TemporaryDocumentTexts
Currently there are places, where `TemporaryDocumentText.objects.get(document=document)` is called (e.g. `documents/utils.py` line 72) but sometimes there exist multiple of such objects so the request fails with `MultipleObjectsReturned`.

Bartzi:
fixes #549


Bartzi:
reworked changes, please check again

PR #554, opened 2017-03-20
add missing items to forbidden urls

### LINKED ISSUE #518: Check for forbidden Urls
We need to check out code that we every URL that should not be used by anyone creating a document is contained in the forbidden url list

Bartzi:
fixes #518


Bartzi:
alirght I've added `ws`...

PR #560, opened 2017-04-10
use kwargs for links in menu items

### LINKED ISSUE #556: fix url param in `minutes:list`
Currently the list of all minutes can be filtered by providing a `GET` parameter to the `minutes:list` url. This parameter is used without a key, which is actually quite confusing and should be avoided.

Bartzi:
fixes #556 by only allowing keyword arguments for links in menu items

### DIFF
--- a/_1327/main/migrations/0013_edit_minutes_menu_item.py
+++ b/_1327/main/migrations/0013_edit_minutes_menu_item.py
@@ -1,0 +1,37 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from django.db import migrations
+
+
+def adapt_links(MenuItem, old_link, new_link, split_str='?'):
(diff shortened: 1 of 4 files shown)


_1327/main/migrations/0013_edit_minutes_menu_item.py
```diff
@@ -0,0 +1,39 @@
...
+		id = menu_item.link.split(split_str)[-1]
+		menu_item.link = '{}{}'.format(new_link, id)
+		menu_item.save()
+
+
+def adapt_minutes_link(apps, schema_editor):
+	# We can't import the MenuItem model directly as it may be a newer
+	# version than this migration expects. We use the historical version.
+	# We can't import the MenuItem model directly as it may be a newer
+	# version than this migration expects. We use the historical version.
```
janno42: thanks for making this clear. reading it twice really helps a lot ;)

PR #580, opened 2017-07-03
Menu item ordering

### LINKED ISSUE #566: Keep menu item order
When a non superuser changes menu item positions, some items might end up in the wrong position. It must be made sure that the change will only affect the items the user can change and all other (potentially invisible) items stay where they are.

Bartzi:
fixes #566


_1327/main/utils.py
```diff
@@ -15,13 +15,28 @@
...
+
+	# check whether we are allowed to change all children of the current level
+	# in case we are not allowed to do that, we have to make sure, that we are not altering the original
+	# order of the menu items, as we are not allowed to do so.
+	all_menu_items_on_this_level = MenuItem.objects.filter(menu_type=MenuItem.MAIN_MENU, parent_id=parent_id)
+	menu_item_order_map = {menu_item: menu_item.order for menu_item in all_menu_items_on_this_level if menu_item.can_edit(user)}
+	if len(menu_item_order_map) != all_menu_items_on_this_level.count():
+		use_old_order = True
+	else:
+		use_old_order = False
```
karyon: `use_old_order = len(menu_item_order_map) != all_menu_items_on_this_level.count()` :)

PR #588, opened 2017-07-19
Minutes creation

### LINKED ISSUE #575: Minutes group selection
We need a better UX for creating minutes when the user is a member of multiple groups who can create minutes.
The suggested way is the following:

1. The "Create Minutes" item in the menu gets subitems for all groups the current user belongs to which can create minutes (if there is more than one, otherwise it stays like it is).
2. After clicking on the respective group, the minutes creation page opens with:
- the title field prefilled with the title last used by the selected group and the possibility to select other previously used titles from a dropdown
- the members of the group initially set as participants
- the same person selected as moderator who was selected in the last minutes document of this group
- the respective group is selected in the field for initial edit rights

janno42:
fix #575 
this makes creating minutes for a specific group easier for users who are in multiple groups that are allowed to create minutes.

### DIFF
--- a/_1327/information_pages/forms.py
+++ b/_1327/information_pages/forms.py
@@ -1,8 +1,3 @@
-from django import forms
-from django.conf import settings
-from django.contrib.auth.models import Group
-from django.core.exceptions import PermissionDenied
-
 from _1327.documents.forms import DocumentForm
 from .models import InformationDocument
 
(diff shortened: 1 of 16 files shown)


_1327/templates/base.html
```diff
@@ -58,15 +58,23 @@
...
-									<li class="indented"><a href="{% url 'documents:create' 'minutesdocument' %}">{% blocktrans count counter=1 %}Minutes{% plural %}Minutes{% endblocktrans %}</a></li>
+									<li><a href="{% url 'documents:create' 'informationdocument'%}">{% trans "Create information page" %}</a></li>
 								{% endif %}
 								{% if CAN_CREATE_POLL %}
-									<li class="indented"><a href="{% url 'documents:create' 'poll' %}">{% trans "Poll" %}</a></li>
+									<li><a href="{% url 'documents:create' 'poll' %}">{% trans "Create poll" %}</a></li>
+								{% endif %}
+								{% if CAN_CREATE_MINUTES %}
+									{% if CAN_CREATE_MINUTES|length > 1 %}
+										<li class="disabled"><a href="#">{% trans "Create minutes" %}</a></li>
```
Bartzi: How about `Create minutes for` instead of `Create minutes`?

_1327/polls/tests.py
```diff
@@ -150,17 +146,16 @@ def test_group_field_hidden_when_user_has_one_group(self):
 		self.assertTrue("Hidden" in str(form.fields['group'][0]))
 
 	def test_group_field_not_hidden_when_user_has_multiple_groups(self):
-		groups = mommy.make(Group, _quantity=2)
-		self.user.groups.add(*groups)
+		other_group = mommy.make(Group)
```
janno42: because it's not the one we were using before. no specific reason.

_1327/templates/base.html
janno42: i would prefer keeping `Create minutes`.

_1327/information_pages/forms.py
```diff
@@ -7,5 +12,30 @@ class Meta(DocumentForm.Meta):
...
 		fields = ['title', 'text', 'url_title', 'is_menu_page', 'comment']
 
+	def __init__(self, *args, **kwargs):
+		user = kwargs.pop('user', None)
+		creation = kwargs.pop('creation', None)
+		super().__init__(*args, **kwargs)
+		staff = Group.objects.get(name=settings.STAFF_GROUP_NAME)
+
+		if user.is_superuser:
+			permitted_groups = Group.objects.filter(permissions__codename="add_informationdocument")
```
janno42: what exactly is "all this"? parts of the init methods of the document types differ, so it's not all the same.

Bartzi: the only difference I can see, is in the filtering of the groups, but maybe I'm wrong :sweat_smile:

janno42: e.g. moderator and participants field

Bartzi: Yeah that's right but you can nevertheless put the duplicated code into the superclass ;)

janno42:
please review again

Bartzi:
:1st_place_medal:

Bartzi:
:1st_place_medal:

PR #629, opened 2018-01-15
add temp prefix for documents

### LINKED ISSUE #618: Don't block URLs with temp documents
Temporary documents block URLs although they are not "really" saved. These documents should be handled differently so that the URL stays available for the first actually saved document that wants to use it.

Bartzi:
fixes #618

### DIFF
--- a/_1327/documents/forms.py
+++ b/_1327/documents/forms.py
@@ -1,8 +1,10 @@
+import re
+
 from django import forms
 from django.conf import settings
 from django.contrib.auth.models import Group, Permission
 from django.contrib.contenttypes.models import ContentType
-from django.core.exceptions import PermissionDenied
+from django.core.exceptions import PermissionDenied, ValidationError
 from django.forms import BaseInlineFormSet
 from django.utils.safestring import mark_safe
 from django.utils.translation import ugettext_lazy as _
@@ -55,9 +57,17 @@ class DocumentForm(forms.ModelForm):
 			self.fields['group'].widget = forms.HiddenInput()
 			self.fields['group'].required = False
 
+		# remove temp prefix from url title in initial data if there is one
+		if self.initial['url_title'].startswith('temp_'):
+			temp_prefix_len = re.search(r'temp_\d+_', self.initial['url_title']).end()
+			self.initial['url_title'] = self.initial['url_title'][temp_prefix_len:]
+
 	def clean_url_title(self):
 		super().clean()
 		url_title = self.cleaned_data['url_title'].lower()
+		if url_title.startswith('temp_'):
+			# the url_title must not start with temp anymore!
+			raise ValidationError(_("You are not allowed to use `temp_` as prefix for the URL"))
 		return slugify_and_clean_url_title(self.instance, url_title)
 
 	@classmethod
--- a/_1327/documents/models.py
+++ b/_1327/documents/models.py
@@ -1,5 +1,7 @@
 from datetime import datetime
 import hashlib
+import re
+
 
 from django.contrib.contenttypes.models import ContentType
 from django.db import models
@@ -107,7 +109,18 @@ class Document(PolymorphicModel):
 
 	@classmethod
 	def generate_default_slug(cls, title):
-		return slugify(title)
+		slug = slugify(title)
+		count = 2
+
+		while Document.objects.filter(url_title=slug).exists():
+			# check whether current title already ends with '_\d+' and handle this case accordingly
+			temp_suffix_match = re.search(r'_(?P<count>\d+)$', slug)
+			if temp_suffix_match:
+				slug = slug[:temp_suffix_match.start()]
+
+			slug = "{}_{}".format(slug, count)
+			count += 1
+		return slug
 
 	@property
 	def view_permission_name(self):
--- a/_1327/documents/tests.py
+++ b/_1327/documents/tests.py
@@ -1,5 +1,6 @@
 from datetime import datetime
 import json
+import re
 import tempfile
 
 from django.conf import settings
@@ -290,7 +291,8 @@ class TestAutosave(WebTest):
 		response = self.app.get(reverse('documents:create', args=['informationdocument']), user=self.user)
 		self.assertEqual(response.status_code, 200)
 		form = response.forms['document-form']
-		url_title = slugify(form.get('title').value)
+		# we need to scrape the url_title from the action as we can not get it from the url_title field in the form
+		url_title = re.search(r'^/(?P<url_title>temp_.+)/', form.action).group('url_title')
 
 		# autosave AUTO
 		response = self.app.post(
@@ -319,7 +321,8 @@ class TestAutosave(WebTest):
 		response = self.app.get(reverse('documents:create', args=['informationdocument']), user=self.user)
 		self.assertEqual(response.status_code, 200)
 		form = response.forms['document-form']
-		url_title2 = slugify(form.get('title').value)
+		# we need to scrape the url_title from the action as we can not get it from the url_title field in the form
+		url_title2 = re.search(r'^/(?P<url_title>temp_.+)/', form.action).group('url_title')
 
 		# autosave second document AUTO
 		response = self.app.post(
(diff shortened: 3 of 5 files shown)


_1327/documents/tests.py
```diff
@@ -472,6 +472,101 @@ def test_autosaves_removed_after_successful_edit(self):
...
+			self.assertIn('temp_', temporary_document_text.document.url_title)
+			temporary_document_text.delete()
+
+	def test_autosave_url_title_prefix_removed_on_save(self):
+		assign_perm("information_pages.add_informationdocument", self.group)
+		assign_perm("minutes.add_minutesdocument", self.group)
+		assign_perm("polls.add_poll", self.group)
+
+		for sub_class in Document.__subclasses__():
+			# 1. create new temporary document
```
janno42: :)

_1327/documents/tests.py
```diff
@@ -472,6 +472,101 @@ def test_autosaves_removed_after_successful_edit(self):
...
+				form.set(field_name, text)
+		form.set('comment', text)
+		form.set('group', self.group.pk)
+		if sub_class == MinutesDocument:
+			participants_field = form.get('participants')
+			participants_field.select_multiple([participants_field.options[0][0]])
+		response = form.submit().follow()
+		self.assertEqual(response.status_code, 200)
+
+	def test_autosave_url_title_prefix_prefix_fallback_solution(self):
```
janno42: :)

_1327/documents/tests.py
```diff
@@ -472,6 +472,101 @@ def test_autosaves_removed_after_successful_edit(self):
...
+		assign_perm("information_pages.add_informationdocument", self.group)
+		assign_perm("minutes.add_minutesdocument", self.group)
+		assign_perm("polls.add_poll", self.group)
+
+		for sub_class in Document.__subclasses__():
+			# 1. create new temporary document
```
janno42: :)

_1327/documents/models.py
```diff
@@ -107,7 +107,12 @@ def generate_new_title(cls):
 
 	@classmethod
 	def generate_default_slug(cls, title):
-		return slugify(title)
+		slug = slugify(title)
+		count = 1
+		while Document.objects.filter(url_title=slug).exists():
+			slug = "{}_{}".format(slug, count)
```
janno42: this would create `test`, `test_1`, `test_2` and so on. please skip `test_1`.

_1327/documents/utils.py
```diff
@@ -51,6 +51,14 @@ def handle_edit(request, document, formset=None, initial=None, creation_group=No
 			# save the document and also save the user and the comment the user added
 			with transaction.atomic(), revisions.create_revision():
 				document.handle_edit(cleaned_data)
+
+				# make sure to remove temp prefix of url_title
+				if document.url_title.startswith('temp_'):
+					document.url_title = document.url_title[len('temp_'):]
+				# check that there is no document already that has that url
+				if Document.objects.filter(url_title=document.url_title).exclude(id=document.id).exists():
+					document.url_title = document.generate_default_slug(document.url_title)
```
janno42: if the `url_title` is something like `test_2` and that title already exists, this would change it to `test_2_2`, correct? maybe you could increment the existing number instead.

_1327/minutes/models.py
```diff
@@ -73,7 +73,7 @@ def generate_new_title(cls):
 	@classmethod
 	def generate_default_slug(cls, title):
 		slug_final = slug = date.today().isoformat()
-		count = 2
+		count = 1
```
janno42: why?

_1327/polls/models.py
```diff
@@ -67,7 +67,7 @@ def get_vote_permission(klass):
 	@classmethod
 	def generate_default_slug(cls, title):
 		slug_final = slug = "new-poll-from-{}".format(date.today().isoformat())
-		count = 2
+		count = 1
```
janno42: why?

_1327/documents/views.py
```diff
@@ -44,7 +44,7 @@ def create(request, document_type):
 		model_class = content_type.model_class()
 		delete_old_empty_pages()
 		title = model_class.generate_new_title()
-		url_title = model_class.generate_default_slug(title)
+		url_title = "temp_{}".format(model_class.generate_default_slug(title))
```
janno42: i would prefer not showing the `temp_` prefix in the form field, because users might always feel the need to remove that. could you maybe change the initial value in the form?

Bartzi:
alright, I did quite some modifications, please review again =)

_1327/documents/forms.py
```diff
@@ -55,6 +57,11 @@ def __init__(self, *args, **kwargs):
 			self.fields['group'].widget = forms.HiddenInput()
 			self.fields['group'].required = False
 
+		# remove temp prefix from url title in initial data if there is one
+		if self.initial['url_title'].startswith('temp_'):
```
janno42: hypothetically, if a user would name the url title `temp_fail` - that would fail, right?

_1327/documents/tests.py
```diff
@@ -246,7 +247,8 @@ def test_autosave_newPage(self):
 		response = self.app.get(reverse('documents:create', args=['informationdocument']), user=self.user)
 		self.assertEqual(response.status_code, 200)
 		form = response.forms['document-form']
-		url_title = slugify(form.get('title').value)
+		# we need to scrape the url_title from the action as we can not get it from the url_title field in the form
+		url_title = re.search(r'^/(?P<url_title>temp_.+)/', form.action).group(1)
```
janno42: you name the group but you don't use the name again. could you remove the name?

janno42: 5 more occurences below

_1327/documents/tests.py
```diff
@@ -324,7 +327,9 @@ def test_create_autosave_non_superuser(self):
 			response = self.app.get(reverse('documents:create', args=[content_type.model.lower()]), user=user)
 			self.assertEqual(response.status_code, 200)
 			form = response.forms['document-form']
-			url_title = slugify(form.get('url_title').value)
+			# we need to scrape the url_title from the action as we can not get it from the url_title field in the form
+			url_title = re.search(r'/(?P<url_title>temp_.+)/', form.action).group(1)
+			url_title = slugify(url_title)
```
janno42: you slugify the title here but not everywhere else. is there a reason for that?

_1327/documents/tests.py
```diff
@@ -473,6 +479,84 @@ def test_autosaves_removed_after_successful_edit(self):
...
+		assign_perm("information_pages.add_informationdocument", self.group)
+		assign_perm("minutes.add_minutesdocument", self.group)
+		assign_perm("polls.add_poll", self.group)
+
+		# create document with appended '_\d+' for testing a special case of url_title handling
+		mommy.make(InformationDocument, url_title=InformationDocument.__name__.lower())
+		test_document = mommy.make(InformationDocument, url_title="{}_2".format(InformationDocument.__name__.lower()))
+
+		self.assertEqual(
+			"{}_3".format(InformationDocument.__name__.lower()),
```
janno42: i find this order confusing but it's not very important. i would prefer `assertEqual(needs_to_be_checked, 42)` because i read "assert that needs_to_be_checked is the expected value 42".

_1327/documents/forms.py
janno42: when does the initial data contain the temp prefix?

_1327/documents/views.py
```diff
@@ -44,7 +44,7 @@ def create(request, document_type):
 		model_class = content_type.model_class()
 		delete_old_empty_pages()
 		title = model_class.generate_new_title()
-		url_title = model_class.generate_default_slug(title)
+		url_title = "temp_{}_{}".format(datetime.utcnow().strftime("%d%m%Y%H%M%S%f"), model_class.generate_default_slug(title))
```
janno42: `max_length` is 255 for `url_title`. this lowers the available number of characters quite a bit. increasing the length might not be necessary, but the forms need to make sure, that the given url title can actually be saved and that the user does not enter a 250 character long title there.

Bartzi: what do you want me to do?
Add constraints to the forms?
Shorten the size of the timestamp?

janno42: adding a constraint to the form sounds good. just throw an error in cleaning if the value is too long.

Bartzi: I just saw, that the form already contains such a constraint :smile:

_1327/documents/forms.py
```diff
@@ -55,9 +57,17 @@ def __init__(self, *args, **kwargs):
...
+			temp_prefix_len = re.search(r'temp_\d+_', self.initial['url_title']).end()
+			self.initial['url_title'] = self.initial['url_title'][temp_prefix_len:]
+
 	def clean_url_title(self):
 		super().clean()
 		url_title = self.cleaned_data['url_title'].lower()
+		if url_title.startswith('temp_'):
+			# the url_title must not start with temp anymore!
+			raise ValidationError(_("You are not allowed to use `_temp` as prefix for the URL"))
```
janno42: `temp_`

PR #630, opened 2018-01-15
fix revision display

### LINKED ISSUE #620: Diff viewer shows wrong versions
The diff viewer for different versions of a document seems to show not the correct differences between two selected versions.
It seems to be one version behind, so that the changes from version 0 to version 1 are shown when selecting the change from 1 to 2. That also makes it impossible to see the changes between the current and the previous version.

Bartzi:
fixes #620

### DIFF
--- a/_1327/documents/tests.py
+++ b/_1327/documents/tests.py
@@ -154,6 +154,50 @@ class TestRevertion(WebTest):
 		self.assertEqual(response.status_code, 200)
 		self.assertIn(reverse('versions', args=[old_url]), response.body.decode('utf-8'))
 
+	def test_version_creation(self):
+		Document.objects.all().delete()
+		self.assertEqual(Document.objects.count(), 0)
+
+		def submit_form(text, form, group, sub_class):
+			for field_name in form.fields.keys():
+				if field_name is not None and 'text' in field_name:
+					form.set(field_name, text)
+			form.set('comment', text)
+			form.set('group', group.pk)
+			if sub_class == MinutesDocument:
+				participants_field = form.get('participants')
+				participants_field.select_multiple([participants_field.options[0][0]])
+			response = form.submit().follow()
+			self.assertEqual(response.status_code, 200)
+
+		group = mommy.make(Group)
+		assign_perm("minutes.add_minutesdocument", group)
+		assign_perm("information_pages.add_informationdocument", group)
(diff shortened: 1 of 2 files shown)


_1327/documents/tests.py
```diff
@@ -153,6 +153,50 @@ def test_revert_to_different_url(self):
 		self.assertEqual(response.status_code, 200)
 		self.assertIn(reverse('versions', args=[old_url]), response.body.decode('utf-8'))
 
+	def test_version_creation(self):
+		Document.objects.get().delete()
```
janno42: shouldn't this be `Document.objects.all().delete()`?

_1327/documents/tests.py
```diff
@@ -153,6 +153,50 @@ def test_revert_to_different_url(self):
...
+			self.assertEqual(response.status_code, 200)
+			form = response.forms['document-form']
+
+			url_title = form.get('url_title').value
+			text_1 = 'something'
+			submit_form(text_1, form, group, sub_class)
+
+			self.assertEqual(Document.objects.count(), expected_count)
+			document = Document.objects.get(url_title=url_title)
+			response = self.app.get(reverse('minutes:edit', args=[document.url_title]), user=self.user)
```
janno42: why `minutes:edit`?

Bartzi:
I've addressed your comments.
Regarding the server:
- we already determined that the most current version of a document is not identical to the current version, which it should be
- so we should look for all documents where the last revision is not identical to the current text and then add a new revision with the current version, that should fix it. But I recommend doing that on a test system

_1327/documents/utils.py
```diff
@@ -49,12 +49,13 @@ def handle_edit(request, document, formset=None, initial=None, creation_group=No
...
 
 			# save the document and also save the user and the comment the user added
-			with transaction.atomic(), revisions.create_revision():
+			with transaction.atomic():
 				document.handle_edit(cleaned_data)
-				document.save()
-				document.save_formset(formset)
-				revisions.set_user(request.user)
-				revisions.set_comment(cleaned_data['comment'])
+				with revisions.create_revision():
```
janno42: can you please add a comment what this does and why `revisions.create_revision()` has to be where it is?

janno42: meaning why it needed to be moved from the line above

Issue #212, opened 2015-11-02
full text search


Nef10:

karyon:
i just came across https://docs.djangoproject.com/en/1.10/ref/contrib/postgres/search/, so there is quite a  bit of functionality already included, but it is rather low-level.

janno42:
#635 introduces a search in minutes documents. This should be extended to a global search also including information documents, polls and document titles.

janno42:
see #658

PR #635, opened 2018-02-05
Search minutes

julkw:
add a search bar in minutes lists with which you ca search through all minutes at once


_1327/settings.py
```diff
@@ -78,6 +78,7 @@
 	'django.contrib.sessions',
 	'django.contrib.messages',
 	'django.contrib.staticfiles',
+	'django.contrib.postgres',
```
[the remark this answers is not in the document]
karyon: this is likely the reason why the tests fail.

it's needed for using the postgres full text search functionality (see [here](https://docs.djangoproject.com/en/2.0/topics/db/search/) and [here](https://docs.djangoproject.com/en/2.0/ref/contrib/postgres/search/), but that's not used right now in this PR as far as i can see.

by using that you would gain stuff like searching for multiple words which would then not need to be adjacent, and weighting the results by some relevancy metric, so it might be worth checking out?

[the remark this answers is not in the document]
janno42: yes, i would suggest that for #212

_1327/minutes/templates/minutes_with_lines_list.html
```diff
@@ -0,0 +1,62 @@
...
+
+					<td style="width: 15%;">
+						{% for label in minute.labels.all %}
+							<span class="label {{ label.class_for_text_color }}" style="background-color: {{ label.color }};">{{ label.title }}</span>
+						{% endfor %}
+					</td>
+					<td style="width: 55%;">
+						<a href="{{ minute.get_view_url }}">{{ minute.title }}</a>
+						{% for line in lines %}
+							<br>{{line}}
```
janno42: `<br />` please

janno42: and `{{ line }}`

_1327/minutes/templates/minutes_list.html
```diff
@@ -19,11 +23,12 @@
 	{% for year, minutes in minutes_list %}
 		<h3 id="year{{ year }}">{{ year }}</h3>
 		<table class="table table-striped">
-			{% for minute in minutes %}
+			{% for minute, lines in minutes %}
 				<tr>
 					<td style="width: 15%;">
 						<a href="{{ minute.get_view_url }}">{{ minute.date | date:"d.m.Y" }}</a>
 					</td>
+
```
janno42: please remove the empty line

_1327/minutes/views.py
```diff
@@ -38,10 +39,61 @@ def list(request, groupid):
 		# we only show documents for which the user has view permissions
 		if not user_checker.has_perm(MinutesDocument.get_view_permission(), m) and (not ip_range_group_name or not ip_range_group_checker.has_perm(MinutesDocument.get_view_permission(), m)):
 			continue
+		permitted_minutes.append(m)
+
+	return permitted_minutes, own_group
+
+
+def search(request, groupid):
+
```
janno42: please remove the empty line

_1327/minutes/views.py
```diff
@@ -38,10 +39,61 @@ def list(request, groupid):
...
+	if(request.method == 'POST'):
+		form = SearchForm(request.POST)
+
```
janno42: please remove the empty line

Issue #638, opened 2018-02-11
Menu item ordering testing error


janno42:
`test_menu_item_ordering` often locally fails for unknown reasons. this can happen even if nothing about menu items has changed in the testing instance.
```
FAIL: test_menu_item_ordering (_1327.main.tests.MenuItemTests)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/vagrant/_1327/main/tests.py", line 357, in test_menu_item_ordering
    self.assertEqual(menu_item.id, int(menu_item_id), 'Menu Item ordering is not as expected')
AssertionError: 10 != 9 : Menu Item ordering is not as expected
```

Bartzi:
This seems to fail only for vagrant users. I suspect this is a problem with Postgres... Would be intersting to see if this fails on production, too...

cmfcmf:
I'm using Python natively on Windows 10: It works fine when installing `psycopg2 == 2.6.2` (the same version as installed on the server) and fails when installing `psycopg2 == 2.7.6.1`.

cmfcmf:
I think this happens because the test orders the menu items strictly by "order", discarding the parent-child relationship.
https://github.com/fsr-de/1327/blob/07f3bb925e2ceca46fc86e839d64aea461e1b0f4/_1327/main/tests.py#L344

Compare this to the code used in the view:

https://github.com/fsr-de/1327/blob/07f3bb925e2ceca46fc86e839d64aea461e1b0f4/_1327/main/views.py#L60-L81

Take this example (itemnames and order):

```
root 5
-- child1 7
---- child2 4
---- child3 5
-- child4 8
```

Correct order: `root, child1, child2, child3, child4`.
Ordering by order: `root, child2, child3, child1, child4`

I think the randomness happens due to the fact that Mommy assigns random order numbers when creating the menu items: https://github.com/fsr-de/1327/blob/07f3bb925e2ceca46fc86e839d64aea461e1b0f4/_1327/main/tests.py#L79-L81

PR #639, opened 2018-02-18
Django 2.0

janno42:
update to django 2.0

something about publishing of minutes seems to be very strange, that needs to be discussed before merging.


_1327/minutes/models.py
```diff
@@ -114,7 +114,7 @@ def show_publish_button(self):
 		return not self.is_in_creation and self.state == MinutesDocument.UNPUBLISHED
 
 	def publish(self, state_id):
-		if state_id not in (MinutesDocument.PUBLISHED, MinutesDocument.PUBLISHED_STUDENT):
+		if state_id in (MinutesDocument.PUBLISHED, MinutesDocument.PUBLISHED_STUDENT):
```
janno42: why did this *ever* work? when testing on the current master i had troubles publishing a minutes document because the server always returned `SuspiciousOperation`. it works on production, but this code didn't change since it's introduction.

janno42:
and it seems like we have database issues, because tests are okay on my testing system. hooray.

_1327/minutes/models.py
Bartzi: `state_id` is a string and not an integer, this is the reason for the failure.

Furthermore, it might make sense to use rename `state_id` to `next_state_id` or `new_state_id`

janno42:
ready for review. the remaining error was not introduced in this pr and will be handled in #638.

janno42:
squashing the poll migrations was necessary to prevent the following error when using sqlite:
`django.db.utils.OperationalError: foreign key mismatch - "polls_choice" referencing "polls_poll"`
i'm not sure why this error occurred in the first place but others seemed to have the same problem in django 2.0 and this was a suggested solution.

janno42:
fixed :)

janno42:
i couldn't find a solution for the sqlite problem so far. it wanted to test whether it works with django 2.1, but django-guardian is not compatible yet.
to not hold this pr any further, there is now #660 dealing with the sqlite problem.

_1327/main/tests.py
```diff
@@ -1,11 +1,13 @@
 import datetime
+from io import StringIO
```
Bartzi: could we clean the imports here? I think this line should be underneath the plain `import` lines, shouldn't it?

janno42: when using
```
import datetime
import json
import re
from io import StringIO
```
the linter says:
```
_1327/main/tests.py:4:1: I100 Import statements are in the wrong order. 'from io import StringIO' should be before 'import re'
```

PR #106, opened 2015-04-13
Fix #96: Add word break for long lines in document diffs

PFischbeck:
This fixes long lines making the table too large.

line-wrapping

PR #641, opened 2018-02-26
Fix overflowing text in content section by forcing word wraps

### LINKED ISSUE #634: Force linebreaks in long text
When displaying a document with sidebar it can happen that a word stretches over the sidebar if it is too long

1327error

LoadingByte:
Fixes #634.

Note: This PR doesn't remove the diff table word wrap fix introduced in #106 since that fix applies to diff tables in general, not just diff tables in content. In case a diff table is used outside of the content div in the future, that fix should still work. Thus, it's not removed and covered by the new general content word wrap policy.

### DIFF
--- a/_1327/static/less/1327.less
+++ b/_1327/static/less/1327.less
@@ -355,9 +355,10 @@ nav li:hover > .dropdown-menu {
 	padding-left: 20px;
 	margin-bottom: 30px;
 	// Wraps all overflowing text in the content section, regardless whether it's part
-	// of the main text, of the attachments list or table or any other content element
-	word-wrap: break-word;
-	word-break: break-all;
+	// of the main text or any other content element.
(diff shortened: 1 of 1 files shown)


_1327/static/less/1327.less
```diff
@@ -354,6 +354,10 @@ nav li:hover > .dropdown-menu {
 .container.content {
 	padding-left: 20px;
 	margin-bottom: 30px;
+	// Wraps all overflowing text in the content section, regardless whether it's part
+	// of the main text, of the attachments list or table or any other content element
+	word-wrap: break-word;
+	word-break: break-all;
```
janno42: this makes the line break always exactly at the end of the container even if a word could entirely be placed onto the next line and would make the break unnecessary (see https://developer.mozilla.org/en-US/docs/Web/CSS/word-break).

PR #646, opened 2018-03-12
Rename python virtualenv in README to reflect virtualenv name in .gitignore

LoadingByte:
The README suggested to name your virtualenv `env`, while the .gitignore was configured to ignore virtualenvs with the name `virtualenv`. The README now also suggests to name your virtualenv `virtualenv`. That's some serious virtualenving for you!


janno42:
I would actually prefer changing the value in the gitignore to `env` because that's shorter to type. would that be okay for you?

LoadingByte:
Yeah, that would, in fact, be my preference to. I'll open another PR for that later.

PR #648, opened 2018-04-23
Emojis for all

### LINKED ISSUE #615: Emoji overview
There should be an easy way to select one of the available emojis when writing a document (e.g. emoji list in a modal).

Bartzi:
fixes #615

### DIFF
--- a/_1327/documents/templates/documents_edit.html
+++ b/_1327/documents/templates/documents_edit.html
@@ -356,6 +356,12 @@
 		}
 	});
 
+	// hotfix emojionearea picker positioning in fullscreen mode
+	$('.emojionearea.md-input').on('keyup', function (event) {
+		var margin = $('.emojionearea-button').css('margin-right');
+		$('.emojionearea-picker').css("right", margin);
+	});
+
 	// handler to update the preview in the modal
 	$('#attachmentModalSelect').on('change', function (event) {
 		var attachmentHash = $(this).val();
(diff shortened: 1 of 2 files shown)


Bartzi:
alright I removed all CDNs please check again =)

janno42:
can we do something against the different renderings in the three views?:

## edit textbox


## preview


## saved page

janno42:
unfortunately the emoji picker currently does not really work in full screen mode (because it's mostly located outside the window then) and it also limits the textbox size to a smaller part of the screen. could you please have a look into this?

Bartzi:
I don't know the exact cause of  the different renderings, but I suspect it is because on Windows emojis might be loaded differently. On Linux I can not see any differences in those three different examples.

I fixed the positioning of the emojiarea in fullscreen mode. Unfortunately, we can not use the autocomplete, that is invoked via typing `:` in fullscreen because we can't change the `z-index` of this autocomplete without rewriting the library, which is not advisable.
But it is still possible to add emojis^^

janno42:
unfortunately still broken. when typing `:` in full screen, the emoji picker moves to the right when opening it again.

Bartzi:
please have a look again ;)

janno42:
the height of the textbox in fullscreen is still too small and the modals for inserting internal links and images is broken.

janno42:
it looks like the modals were a caching error on my site, so please disregard

janno42:
can you please add:
```
.emojionearea md-input {
    height: 450px;
}
.emojionearea-editor {
    height: 100%;
    max-height: none;
}
```

Bartzi:
Please check again, I found another bug in fullscreen mode...
when pressing repeatedly on one emoji, the picker went to the right and out of the window. This should now be fixed.

PR #653, opened 2018-06-18
copy shortlink url to clipboard on click

### LINKED ISSUE #651: Add copy to clipboard button for shortlinks
After I've created a shortlink I want to be able to copy it into my clipboard with a single click on button, or the link itself. This way there is no need for me to retype parts of the link myself, when I want to show it somewhere

Bartzi:
fixes #651

### DIFF
--- a/_1327/shortlinks/templates/shortlinks_index.html
+++ b/_1327/shortlinks/templates/shortlinks_index.html
@@ -15,16 +15,20 @@
 				<th class="col-xs-2">{% trans "URL" %}</th>
 				<th class="col-xs-3">{% trans "Link" %}</th>
 				<th class="col-xs-2">{% trans "Created" %}</th>
-				<th class="col-xs-2">{% trans "Visit count" %}</th>
+				<th class="col-xs-1">{% trans "Visitors" %}</th>
 				<th class="col-xs-2">{% trans "Last access" %}</th>
-				<th class="col-xs-1"></th>
+				<th class="col-xs-2"></th>
 			</tr>
 		</thead>
 		{% for shortlink in shortlinks %}
 			<tr>
 				<td>
 					<input type="hidden" class="shortlinkinput" name="shortlink" value="{{ shortlink.id }}">
-					{{ shortlink.url_title }}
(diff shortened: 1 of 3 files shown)


Bartzi:
OKay, I changed the behaviour of the implementation. It works now with multiple links and looks different. I don't know how to make it nicer, please help if you don't like it now^^

_1327/shortlinks/templates/shortlinks_index.html
```diff
@@ -15,16 +15,19 @@
 				<th class="col-xs-2">{% trans "URL" %}</th>
 				<th class="col-xs-3">{% trans "Link" %}</th>
 				<th class="col-xs-2">{% trans "Created" %}</th>
-				<th class="col-xs-2">{% trans "Visit count" %}</th>
+				<th class="col-xs-1">{% trans "Visitors" %}</th>
 				<th class="col-xs-2">{% trans "Last access" %}</th>
-				<th class="col-xs-1"></th>
+				<th class="col-xs-2"></th>
```
janno42: do the buttons need more space in your browser?

_1327/shortlinks/templates/shortlinks_index.html
```diff
@@ -78,5 +81,38 @@
...
+		shortlinkURLButton.on('click', function (event) {
+			var urlHolder = $(this).parent().find(".shortlink-urltitle");
+			var baseURL = urlHolder.html();
+			var fullURl = document.location.origin + '/' + baseURL;
+
+			// create dummy textarea, paste the text to copy and copy that text
+			var el = document.createElement('textarea');
+			el.value = fullURl;
+			el.setAttribute('readonly', '');
+			el.style.position = 'absolute';
```
janno42: why do you need the positioning - can't you just hide the element? i would actually prefer adding a hidden element on the page with the full url in the template and copying from there.

Bartzi: we need the positioning, because otherwise the display might flicker shortly, once we add the element. We can not add a hidden element with the full url, because we don't know the full url on page creation, that's why we need Javascript to do that.

PR #659, opened 2018-08-20
Delete autosaves

### LINKED ISSUE #606: Enable deleting autosaves
There should be an option to delete an autosaved document after restoring and opening it.

Bartzi:
Allows the deletion of autosaves, check it out ;)
fixes #606

### DIFF
--- a/_1327/documents/templates/documents_edit.html
+++ b/_1327/documents/templates/documents_edit.html
@@ -35,6 +35,11 @@
 				{% trans 'Delete' %}
 			</button>
 		{% endif %}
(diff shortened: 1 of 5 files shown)


cmfcmf:
> It seems like there can only be one autosave per document.

A few days ago I remember having a document with two autosaves.

cmfcmf:
>> It seems like there can only be one autosave per document.
> 
> A few days ago I remember having a document with two autosaves.

I was wrong, I saw multiple autosaves when creating a new document (thanks @janno42).

PR #663, opened 2018-09-03
redirect users to same page after they login on their own initiative

### LINKED ISSUE #604: redirect after login
ref #434 

when i am on any page and not logged in, going to the login page and logging in should redirect me to the page i was on before.

Bartzi:
fixes #604

### DIFF
--- a/_1327/minutes/templates/minutes_list.html
+++ b/_1327/minutes/templates/minutes_list.html
@@ -1,5 +1,6 @@
 {% extends 'base_with_sidebar.html' %}
 {% load i18n %}
+{% load redirect_login_tags %}
 
 {% block title %}
 	{% blocktrans count counter=2 %}Minutes{% plural %}Minutes{% endblocktrans %}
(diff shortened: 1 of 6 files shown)


_1327/minutes/templates/minutes_list.html
```diff
@@ -63,15 +64,16 @@ <h3 id="year{{ year }}">{{ year }}</h3>
...
-	            <em>
-	                {% blocktrans with anchor='<a href="'|add:anchor_url|add:'">'|safe anchor_end='</a>'|safe %}You might have to {{ anchor }} login {{ anchor_end }} first.{% endblocktrans %}
-	            </em>
-	        {% endif %}
-	    {% endblock %}
+	{% empty %}
+		{% block minutesempty %}
+			<em>{% trans "No minutes available." %}</em>
+			{% url "login" as anchor_url %}
+			{% redirect_login as redirect_params %}
```
janno42: why can't you use it directly?

Bartzi: I think this would break the translation, wouldn't it?

janno42: `{% blocktrans with anchor='<a href="'|add:anchor_url|add:redirect_login|add:'">'|safe anchor_end='</a>'|safe %}...{% endblocktrans %}` should work fine i think.

Bartzi: I think the problem is that we can not nest multiple template filters and the template system does not necessarily know that it should render the filter,but it would make sense to try it

PR #669, opened 2018-11-06
Menu item index

janno42:
fix menu item visibility in menu item index.
currently items are also visible if the user can edit any other item on the same level.


Bartzi:
Hmm, interesting. Could you please add a test that makes sure your change helps with this problem?

PR #670, opened 2018-11-12
Add Yarn for managing static files

### LINKED ISSUE #668: Markdown falsely interpreted as emojis
In some cases markdown text is interpreted as emojis in the text editor. The markdown editor should not replace `#` and `*` (and maybe other characters as well) by anything else.

----

Entered text:
```
# TOP 1: Test
- lorem ipsum
- dolor sit amet *consectetur*
```

After reloading this is shown in the editor text field:
markdownemoji

The rendered result however is correct, even after saving the document again, so it's only a problem in the editor, not in the saved data:
markdownresult

Bartzi:
This PR adds static file management via the javascript package manager [Yarn](https://yarnpkg.com/).
This has the advantage that it is easy to update our javascript requirements and also have automatic dependency management. We also do not need to have all javascript libraries in our repository.

This also fixes #668 

Please check all pages whether they still work, as we do not have automated fronted tests yet, we can not do this automatically. I tried to do nearly everything, but I'm sure I missed something ;)


cmfcmf:
Maybe you have these on your TODO list already, just in case you don't:
- the [`.gitmodules` file](https://github.com/fsr-itse/1327/blob/07f3bb925e2ceca46fc86e839d64aea461e1b0f4/.gitmodules#L1-L9) should be deleted and
- the README adjusted accordingly (no longer requiring installation of submodules) and
- the Vagrant scripts be adjusted 

Nice work! 👍 
And 💯 for choosing Yarn over npm.

deployment/provision_vagrant_vm.sh
```diff
@@ -34,3 +41,8 @@ git submodule update --init
 sudo -H -u vagrant python3 manage.py migrate --noinput
 sudo -H -u vagrant python3 manage.py collectstatic --noinput
 sudo -H -u vagrant python3 manage.py compilemessages
+
+
+# setup static files
+cd /vagrant/_1327/static
+sudo -H -u vagrant yarn
```
janno42:
```suggestion
sudo -H -u vagrant yarn --no-bin-links
```
this should fix the problem using a vagrant vm on windows

_1327/static/less/diffview.less
```diff
@@ -1,97 +1,49 @@
...
+	border: 1px solid @gray;
+	white-space: pre-wrap;
 	display: table;
 	margin-left: auto;
 	margin-right: auto;
 	min-width: 100%;
 }
-table.diff tbody { 
-	font-family:Courier, monospace
-}
```
janno42: this should not be removed, we want to keep monospace font for the diff view

_1327/static/less/diffview.less
```diff
@@ -1,97 +1,49 @@
...
-	font-family:verdana,arial,'Bitstream Vera Sans',helvetica,sans-serif;
-	background:@gray-lighter;
-	font-size:11px;
-	font-weight:normal;
-	border:1px solid @gray;
-	color:@gray-dark;
-	padding:.3em .5em .1em 2em;
-	text-align:right;
-	vertical-align:top
+	font-family: verdana, arial, 'Bitstream Vera Sans', helvetica, sans-serif;
```
janno42: here are some additional changes that shouldn't be removed.

_1327/documents/templates/documents_edit.html
```diff
@@ -372,7 +372,7 @@ <h4 class="modal-title" id="deleteAutosaveHeader">{% trans "Do you really want t
 	// init emoji stuff
 	$('#id_text').emojioneArea({
 		shortnames: true,
-		saveEmojisAs: 'unicode',
+		saveEmojisAs: 'shortname',
```
janno42: the diff view doesn't display emojis anymore when the document is edited after this change because they are no longer stored in unicode. why does this help again?

Bartzi: Hmm, I hoped this makes it nicer when writing the text, as emojis do not necessarily get translated into the unicode representation directly in the editor...  but if it breaks other things... we should keep it the way it was ;)

deployment/provision_vagrant_vm.sh
```diff
@@ -34,3 +41,8 @@ git submodule update --init
 sudo -H -u vagrant python3 manage.py migrate --noinput
 sudo -H -u vagrant python3 manage.py collectstatic --noinput
 sudo -H -u vagrant python3 manage.py compilemessages
+
+
+# setup static files
+cd /vagrant/_1327/static
+sudo -H -u vagrant yarn --no-bin-links
```
cmfcmf: Setting up static files should probably happen before calling `collectstatic`?

PR #672, opened 2018-11-26
use monospace font in editor

Bartzi:
our editor now uses a monospace font :tada:


janno42:
i think we will keep the usual sans serif font in the editor. i hope you're okay with that :)

PR #685, opened 2019-02-11
Markdown fixes

### LINKED ISSUE #681: Inserting internal links is broken
When inserting internal links in the markdown editor, the link is added at the end of the document. It should be added at the current cursor position.

Bartzi:
fixes #681 

It is now possible to use all markdown editor related buttons and we also have the original text editor back :tada:, but it is a bit hacky now. Please test thoroughly!

### DIFF
--- a/_1327/documents/templates/documents_edit.html
+++ b/_1327/documents/templates/documents_edit.html
@@ -350,10 +350,10 @@
 							// find the right spot to insert the text and create the correct markdown
 							var markdown = $('#selectInternalLinkModal').val();
 							if (markdown !== null) {
-								linkText = markdown.match(/\[(.*?)\]/)[1]; // we want the text inside the [ ]
-								selected = e.getSelection();
+								var linkText = markdown.match(/\[(.*?)\]/)[1]; // we want the text inside the [ ]
+								var selected = e.getSelection();
 								e.replaceSelection(markdown);
-								cursor = selected.start;
+								var cursor = selected.start;
 								e.setSelection(cursor + 1, cursor + linkText.length + 1); // +1 because we do not want to select the starting [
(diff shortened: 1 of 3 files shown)


_1327/documents/templates/documents_edit.html
```diff
@@ -365,32 +365,59 @@ <h4 class="modal-title" id="deleteAutosaveHeader">{% trans "Do you really want t
...
+					callback: function(e) {
+						// check if we already created an emojionearea object
+						// if we did, we can just show it
+						var emojiHolder = $('.emojioneholder');
+						if (emojiHolder.length) {
+							emojiHolder.data("emojioneArea").showPicker();
+							return;
+						}
+
+						$(".btn-toolbar").append("<div class='emojioneholder'>hello</div>");
```
janno42: hello too :)

_1327/static/js/utils.js
```diff
@@ -48,3 +48,24 @@ var debounce = function (func, threshold) {
...
+     * The splice() method changes the content of a string by removing a range of
+     * characters and/or adding new characters.
+     *
+     * @this {String}
+     * @param {number} start Index at which to start changing the string.
+     * @param {number} delCount An integer indicating the number of old chars to remove.
+     * @param {string} newSubStr The String that is spliced in.
+     * @return {string} A new string with the spliced substring.
+     */
+    String.prototype.splice = function(start, delCount, newSubStr) {
```
janno42: this is used in only one place and `delCount` is `0`, so it looks like too many lines of code (or rather documentation) for what's happening. why not simply use the substring from the beginning to the cursor position, then the added text and finally the substring from the cursor position to the end in `documents_edit.html`?

PR #686, opened 2019-03-19
Make logo and favicon customizable

janno42:


_1327/settings.py
```diff
@@ -193,6 +193,9 @@
 
 STATIC_URL = '/static/'
 
+LOGO_FILE = ""
+FAVICON_FILE = "/static/images/favicon.ico"
```
Bartzi: I don't think it is a very great idea to hardcode the path to that static file here. Expecially the `static url` could make a problem. I think we have two solutions:

1. use  `FAVICON_FILE = STATIC_URL + "images/favicon.ico"`, or
2. use the static template processor in our templates like so: `{% static "{{FAVICON_FILE}}" %}` (if s.th. like that can work)

We should do this because those files should always reside in the static folder (since they are static files) and it lets us keep the flexibility to easily change the `STATIC_URL` if we ever want to do that... I personally favour option 2, but I'm not sure whether that can work...

janno42: the second option does not work.
please review again.

janno42:
a ui test whether the logo is shown? i wouldn't say its necessary and we don't test anything alike yet but if you insist, i can add it.

Bartzi:
Well, I think we want to have a better test coverage for our system, so this seems to be a good opportunity :wink: 
I insist on a test, please add one.

PR #701, opened 2019-07-21
Navbar Cache

### LINKED ISSUE #683: Cache navigation
The navigation bar should be cached to dramatically increase page load times.

felixrindt:
Fixes #683 

Still missing profiling


_1327/main/tests.py
```diff
@@ -372,6 +426,9 @@ def test_menu_item_visible_for_user(self):
 		self.sub_sub_item.save()
 		assign_perm(self.sub_sub_item.view_permission_name, self.user, self.sub_sub_item)
 
+		# TODO: Still missing a cache clear when group permissions change via django-guardian
+		# delete_navbar_cache_for_users([self.user])
```
felixrindt: I didn't find a way to hook into django-guarding to do this nicely, yet :/

_1327/main/utils.py
```diff
@@ -16,6 +19,19 @@
...
+	try:
+		for user in users:
+			key = make_template_fragment_key('navbar', [user.username, 'de'])
+			cache.delete(key)
+			key = make_template_fragment_key('navbar', [user.username, 'en'])
+			cache.delete(key)
+	except OperationalError:  # During migration for the test db, the cache table might not exist yet
+		from _1327.settings import TESTING
+		if not TESTING:
+			raise
```
felixrindt: The tests fail during the migration for the test database.

_1327/settings.py
```diff
@@ -26,6 +26,9 @@
 
 # SECURITY WARNING: don't run with debug turned on in production!
 DEBUG = True
+ENABLE_DEBUG_TOOLBAR = False
```
janno42: please move to own commit and finish the implementation :)

_1327/settings.py
```diff
@@ -265,17 +275,33 @@
...
 
 # Create a localsettings.py to override settings per machine or user, e.g. for
 # development or different settings in deployments using multiple servers.
 _LOCAL_SETTINGS_FILENAME = os.path.join(BASE_DIR, "localsettings.py")
 if os.path.exists(_LOCAL_SETTINGS_FILENAME):
 	exec(compile(open(_LOCAL_SETTINGS_FILENAME, "rb").read(), _LOCAL_SETTINGS_FILENAME, 'exec'))
 del _LOCAL_SETTINGS_FILENAME
+
+
+# Django debug toolbar settings
```
janno42: see above

_1327/user_management/models.py
```diff
@@ -81,3 +82,22 @@ def add_to_default_group(sender, **kwargs):
 		user = kwargs.get('instance')
 		group, __ = Group.objects.get_or_create(name=settings.DEFAULT_USER_GROUP_NAME)
 		user.groups.add(group)
+
+
+@receiver(m2m_changed, sender=UserProfile.groups.through)
+def group_user_changed(instance, action, reverse, pk_set, **kwargs):
+	if 'post' in action:
+		if not reverse:  # A user's group was changed
```
janno42: `A user's groups were changed`

_1327/settings.py
felixrindt: I picked these changes away for now :)

felixrindt:
I'm closing this, because the scope has changed. Caching the navbar HTML is not helpful, because it's context preprocessors still need to run to enable the breadcrumbs.

What we looked at is that `mark_selected` works using recursion, doing a database query on every call, which is *probably* why creating the navbar is slow. 

We could get rid of this recursion and avoid the search by going up the hierarchy from the currently selected element. @janno42 might do this in a seperate PR.

PR #702, opened 2019-07-21
introduced group admin view

### LINKED ISSUE #647: Failing group membership changes
Changing which users belong to a group in the respective admin menu fails with a `400` server return message.

jeriox:
closes #647


_1327/user_management/forms.py
```diff
@@ -47,3 +49,25 @@ class UserImpersonationForm(forms.Form):
...
+	add_minutes = BooleanField(required=False)
+	add_poll = BooleanField(required=False)
+	users = ModelMultipleChoiceField(UserProfile.objects.all())
+
+	class Meta:
+		model = Group
+		fields = ("name",)
+
+	def __init__(self, *args, **kwargs):
+		super().__init__(**kwargs)
```
janno42: you need to replace this with ```super(GroupEdit, self).__init__(*args, **kwargs)```, then the form should be valid on submission.

_1327/user_management/forms.py
```diff
@@ -47,3 +49,25 @@ class UserImpersonationForm(forms.Form):
 	def __init__(self, *args, **kwargs):
 		super(UserImpersonationForm, self).__init__(*args, **kwargs)
 		self.fields['username'].widget.attrs['class'] = 'select2-selection'
+
+
+class GroupEdit(forms.ModelForm):
```
janno42: please rename the form to `GroupEditForm` to make clear that it is a form :)

jeriox:
@janno42 Apparently it is not possible to provide inital values to custom Fields on a ModelForm from the model. Thus I re-initialize the fields in the init method. Can you think of a better way for that?

janno42:
setting the values in the init method is perfectly fine.

PR #703, opened 2019-08-05
Debugtoolbar

felixrindt:
and also factor out the navbar template...


hendraet:
I have no idea if all of this is correct and a good way of doing it :D

PR #707, opened 2019-11-04
remove dependency on globally installed less

Pottiman:
This pull request moves the installation of `less` to the `yarn` install step.
The static precompiler is told to use the `lessc` in `BASE_DIR/static/node_modules/.bin`.

The corresponding travis install step and vagrant provisioning step were removed as well.

### DIFF
--- a/deployment/provision_vagrant_vm.sh
+++ b/deployment/provision_vagrant_vm.sh
@@ -13,8 +13,6 @@ echo "deb https://dl.yarnpkg.com/debian/ stable main" | sudo tee /etc/apt/source
 apt-get update
 apt-get -q install -y yarn
 
-yarn global add less
-
 # setup postgres
 apt-get -q install -y postgresql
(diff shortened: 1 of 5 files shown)


deployment/provision_vagrant_vm.sh
```diff
@@ -11,7 +11,6 @@ curl -sS https://dl.yarnpkg.com/debian/pubkey.gpg | sudo apt-key add -
 echo "deb https://dl.yarnpkg.com/debian/ stable main" | sudo tee /etc/apt/sources.list.d/yarn.list
 
 apt-get update
-apt-get -q install -y yarn
```
janno42: why do you remove this?

Pottiman: You're right, i got the wrong line, should be fixed now.

jeriox:
@Pottiman I am removing less from this project in #722

PR #716, opened 2020-03-30
OpenID login

janno42:
see https://github.com/e-valuation/EvaP/pull/1366 for reference

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,6 +32,8 @@ At that point, one created a vagrant box, running a [PostgreSQL](https://www.pos
 ./manage.py createsuperuser --username=root
 ```
 
+To login with your local user instead of the default OpenID login, you have to visit `http://localhost:8000/local_login=1`
+
 ### Virtual Environment
 
 Another way of executing this django application is the use of a virtual python environment. This way bypasses the needs for a virtual machine and simplifies the life with multiple python versions installed:
--- a/README.md
+++ b/README.md
@@ -32,7 +32,7 @@ At that point, one created a vagrant box, running a [PostgreSQL](https://www.pos
 ./manage.py createsuperuser --username=root
 ```
 
-To login with your local user instead of the default OpenID login, you have to visit `http://localhost:8000/local_login=1`
+To login with your local user instead of the default OpenID login, you have to visit `http://localhost:8000/login?local_login=1`
 
 ### Virtual Environment
 
--- a/_1327/settings.py
+++ b/_1327/settings.py
@@ -52,7 +52,7 @@ FORBIDDEN_URLS = [
 	"admin", "login", "logout", "documents", "information_pages", "minutes", "polls", "list", "view_as", "abbreviation_explanation",
 	"menu_items", "menu_item_delete", "menu_item", "create", "edit", "delete", "update_order", "hijack", "unlinked", "revert", "search", "download",
 	"update", "attachment", "no-direct-download", "autosave", "publish", "render", "delete-cascade", "versions", "permissions", "attachments",
-	"shortlink", "shortlinks", "preview", "get", "change", "autosave", "ws",
+	"shortlink", "shortlinks", "preview", "get", "change", "autosave", "ws", "oidc",
 ]
 
 ANONYMOUS_GROUP_NAME = "Anonymous"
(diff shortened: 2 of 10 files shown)


_1327/user_management/templates/login.html
```diff
@@ -12,18 +12,29 @@
...
+		<p>
+			{% trans 'Log in using HPI OpenID Connect.' %}
+		</p>
+		{% buttons %}
+			<a href="{% url 'oidc_authentication_init' %}{% if 'next' in request.GET %}?next={{ request.GET.next }}{% endif %}" class="btn btn-primary">
+				{% trans 'Login' %}
+			</a>
+		{% endbuttons %}
+	{% else %}
+		<form action="{% url 'login' %}" method="post" class="form-horizontal" role="form">
```
jeriox, line 25: as this posts to the plain login route, the user is shown the openid login screen instead of a form with errors after entering wrong user credentials

_1327/settings.py
```diff
@@ -270,6 +278,25 @@
 # Set this to the ID of the document that shall be shown as Main Page
 MAIN_PAGE_ID = -1
 
+
+# OpenID Login
+# replace 'example.com', OIDC_RP_CLIENT_ID and OIDC_RP_CLIENT_SECRET with real values in localsettings when activating
+ACTIVATE_OPEN_ID_LOGIN = False
+OIDC_RENEW_ID_TOKEN_EXPIRY_SECONDS = 60 * 60 * 24 * 7  # one week
```
Bartzi, line 285: how about `datetime.timedelta(weeks=1).total_seconds()`?

README.md
```diff
@@ -32,6 +32,8 @@ At that point, one created a vagrant box, running a [PostgreSQL](https://www.pos
 ./manage.py createsuperuser --username=root
 ```
 
+To login with your local user instead of the default OpenID login, you have to visit `http://localhost:8000/local_login=1`
```
Bartzi, line 35: this should be `http://localhost:8000/login?local_login=1`

_1327/user_management/templates/login.html
Bartzi: yes, I can confirm that problem.

janno42: i can think of only one use case where this might be a problem: when the admin user wants to login on production and provides wrong credentials. all other users will use open id on production, and in development open id will be deactivated.
so i don't think we need to add additional code for this.

Bartzi: you might be right, but this is still some kind of a bug and it should be handled correctly. That means that even an admin user should see the login form and the system should tell him that has been a problem logging him in, because username and password did not match

README.md
```diff
@@ -32,7 +32,7 @@ At that point, one created a vagrant box, running a [PostgreSQL](https://www.pos
 ./manage.py createsuperuser --username=root
 ```
 
-To login with your local user instead of the default OpenID login, you have to visit `http://localhost:8000/local_login=1`
+To login with your local user instead of the default OpenID login, you have to visit `http://localhost:8000?local_login=1`
```
Bartzi, line 35: you are still missing the `/login` in your URL :wink:

janno42:
about the tests:

janno42:
about the tests:
1) the local login form is tested in `user_management.tests.UsecaseTests`
2) mocking the openid login is not in scope for this pr. if someone feels like this is really needed, we have to open an issue for that :)

README.md
janno42: oh my...

Bartzi:
Yes, right :sweat_smile: the local login is already tested.
But it would make sense to check that changing the openid setting actually changes the behaviour of the page the way it is intended to be. This is just one small test :wink:.
I think you are right about the OpenID mocking, please open a new issue for that.

_1327/user_management/templates/login.html
janno42: if anyone wants to look into this in the next days - feel welcome. it would require some logic to know which login was used and errors then need to be handled accordingly.
as this pr is blocking our much needed production update and i currently don't want to invest more time into the project than absolutely necessary, i won't implement it myself.

janno42: see #721 for this issue

PR #726, opened 2020-04-30
Minutes editor textblocks buttons

### LINKED ISSUE #623: Editor buttons for minutes text blocks
The editor should get buttons which add the specific text blocks available for minutes documents.
The buttons for begin and end time should also automatically insert the current time.

gritfessel:
closes #623

### DIFF
--- a/_1327/documents/templates/documents_edit.html
+++ b/_1327/documents/templates/documents_edit.html
@@ -649,14 +649,22 @@
 								const oldSelectionStartPosition = markdownEditor[0].selectionStart;
 								const oldSelectionEndPosition = markdownEditor[0].selectionEnd;
 								
+								// set cursor position to absolute insert position
 								markdownEditor[0].selectionStart = unfinishedBreakPosition + relativeInsertPosition;
 								markdownEditor[0].selectionEnd = unfinishedBreakPosition + relativeInsertPosition;
-								
+								const currentSelectionStartPosition = markdownEditor[0].selectionStart;
+
 								textToInsert = currentTime;
 								insertText(textToInsert, markdownEditor);
-								
-								markdownEditor[0].selectionStart = oldSelectionStartPosition; // cursor stays at its old position
+
+								// cursor shall stay at its old (relative) position
+								markdownEditor[0].selectionStart = oldSelectionStartPosition;
 								markdownEditor[0].selectionEnd = oldSelectionEndPosition;
+								if (oldSelectionStartPosition > currentSelectionStartPosition) {	
+									markdownEditor[0].selectionStart += textToInsert.length;
+									markdownEditor[0].selectionEnd += textToInsert.length;
+								}
+
 							} else if (startedMeeting < 0) { // meeting only needs to be started once
 								textToInsert = "|start|(" + currentTime + ")";
 								insertText(textToInsert, markdownEditor);
--- a/_1327/documents/templates/documents_edit.html
+++ b/_1327/documents/templates/documents_edit.html
@@ -631,30 +631,56 @@
                     data: [{
                         name: "cmdAddStartText",
                         toggle: false,
-                        title: "Add start text",
+                        title: "Begin or continue the meeting",
                         icon: "glyphicon glyphicon-play",
 						callback: function (e) {
                             const markdownEditor = e.$editor.find(".md-input");
 
 							const currentTime = getCurrentTime();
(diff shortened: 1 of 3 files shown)


_1327/documents/templates/documents_edit.html
```diff
@@ -403,6 +403,26 @@ <h4 class="modal-title" id="deleteAutosaveHeader">{% trans "Do you really want t
...
+			const currentHour = currentDate.getHours().toString();
+			let currentMinute = currentDate.getMinutes().toString();
+			if(currentMinute.length == 1) currentMinute = "0" + currentMinute;
+
+			return currentHour + ":" + currentMinute;
+		}
+
+		function insertText(textToInsert, markdownEditor) {
+			const markdownText = markdownEditor[0].value;
+			const cursorPosition = markdownEditor[0].selectionStart;
```
marcelgarus, line 417: Should the insertion position always be the cursor position? One could argue that the `|start|(...)` text should always be placed at the beginning, while the `|end|(...)` should always be inserted at the end, even when the user is still changing some stuff in the middle.
Then again, I'm not sure how deterministic this feels for users. Especially, I'm not sure how inserting text into text fields and changing its selection affects the scrolling position. If the text field is rather small and users hit the end button repeatedly without realizing new text gets inserted somewhere they can't see it, that would certainly be worse than the current implementation.

_1327/documents/templates/documents_edit.html
```diff
@@ -403,6 +403,26 @@ <h4 class="modal-title" id="deleteAutosaveHeader">{% trans "Do you really want t
 		// add datepicker to date inputs
 		$("input[name*='date']:not([readonly='True'])").datepicker( $.datepicker.regional["{{ LANGUAGE_CODE }}"] );
 
+		function getCurrentTime() {
+			const currentDate = new Date();
+			const currentHour = currentDate.getHours().toString();
+			let currentMinute = currentDate.getMinutes().toString();
+			if(currentMinute.length == 1) currentMinute = "0" + currentMinute;
```
marcelgarus, line 410: This is certainly nitpicking, but I prefer putting a space between control structures and their parenthesis to distinguish them from function calls. Quickly looking over other parts of this file, it seems like this is the preferred formatting in this project.

_1327/documents/templates/documents_edit.html
```diff
@@ -602,7 +622,125 @@ <h4 class="modal-title" id="deleteAutosaveHeader">{% trans "Do you really want t
...
+						callback: function (e) {
+                            const markdownEditor = e.$editor.find(".md-input");
+
+							const currentTime = getCurrentTime();
+							let textToInsert;
+
+							const unfinishedBreak = markdownEditor[0].value.match(/\|break\|\((\d+):(\d+)\)\(\)/);
+							const startedMeeting = markdownEditor[0].value.search(/\|start\|\((\d+):(\d+)\)/);
+
+							if(unfinishedBreak) {
```
marcelgarus, line 642: see above

_1327/documents/templates/documents_edit.html
```diff
@@ -602,7 +622,125 @@ <h4 class="modal-title" id="deleteAutosaveHeader">{% trans "Do you really want t
...
+					},
+					{
+                        name: "cmdAddEndText",
+                        toggle: false,
+                        title: "End the meeting",
+                        icon: "glyphicon glyphicon-stop",
+						callback: function (e) {
+                            const markdownEditor = e.$editor.find(".md-input");
+							
+							const currentTime = getCurrentTime();
```
marcelgarus, line 693: The indentation is weird here as well as in some other places in this file. I'm pretty sure there's an easy way to enable auto-formatting in most editors.
For example, in VS Code on Windows, in the settings (<kbd>Ctrl</kbd> + <kbd>,</kbd>) you can enable "Editor: Format On Save" (although you probably need to have a HTML/JS language extension installed for that to work).

_1327/documents/templates/documents_edit.html
janno42: both should be placed at the cursor position. there might be some introduction before the beginning time and similarly there can be text after the ending mark.

_1327/documents/templates/documents_edit.html
janno42, line 410: please add a leading zero for the hours as well. i would have preferred proper string formatting, but it seems like javascript is not really good with that.

T4rikA (approved):

PR #731, opened 2020-04-30
Nicer readme

felixrindt:


README.md
```diff
@@ -20,18 +20,24 @@ Freshly created code needs to be tested - besides our use of unit tests, linting
...
 
-At that point, one created a vagrant box, running a [PostgreSQL](https://www.postgresql.org/) database server, [Apache](https://httpd.apache.org/) web server and the [Django](https://www.djangoproject.com/) application. The contents are available on the default port `8000`, which allows one to access the website at `http://localhost:8000`. To create a new superuser for the application, execute `vagrant ssh` in the project directory and trigger the django user management system:
+This will set up a virtual machine and run it. Running this for the first time might take a while.
+
+To connect to it and start the application do:
 
 ```bash
-./manage.py createsuperuser --username=root
+vagrant ssh
+# This will take one inside the virtual machine
```
janno42, line 35: i read this sentence multiple times to understand it. in other places you replaced "one" with "you". that would help a lot here as well :)

deployment/provision_vagrant_vm.sh
```diff
@@ -46,5 +45,6 @@ sudo -H -u vagrant yarn --no-bin-links
 # setup 1327
 cd /vagrant
 sudo -H -u vagrant python3 manage.py migrate --noinput
+sudo -H -u vagrant python3 manage.py shell -c "from django.contrib.auth import get_user_model; User = get_user_model(); User.objects.create_superuser('admin', email='admin@localhost', password='admin'); print('Created superuser')"
```
janno42, line 48: you could make this more explicit as we know what our user model is (`UserProfile`).
also, please use a different email address with explicitly specified tld, so that the user form will accept changes to the user profile later on without form field errors.
it would also be great if you could add the admin user to a group and give this group the three add permissions for minutes, polls and information documents.

README.md
felixrindt: ah yea I forgot to replace the `one` here :)

_1327/main/management/commands/add_superuser.py
```diff
@@ -0,0 +1,37 @@
...
+            self.stdout.write("DEBUG is disabled. Are you sure you are not running")
+            if input("on a production system and want to continue? (yes/no)") != "yes":
+                self.stdout.write("Aborting...")
+                return
+
+        self.stdout.write('Adding superuser admin with password admin.')
+
+        UserProfile.objects.filter(username="admin").delete()
+        user = UserProfile(username='admin', email='admin@example.com', password='admin', is_superuser=True)
+        user.save()
```
felixrindt, line 23: This doesn't seem to actually create the user. They are not available when using `./manage.py run`. What do I need to do? I used `create_superuser` in earlier commits.

felixrindt: works now but i dont know why

_1327/main/management/commands/add_superuser.py
```diff
@@ -0,0 +1,36 @@
...
+                self.stdout.write("Aborting...")
+                return
+
+        self.stdout.write('Adding superuser admin with password admin.')
+
+        UserProfile.objects.filter(username="admin").delete()
+        user = UserProfile.objects.create_superuser(username='admin', email='admin@example.com', password='admin')
+        user.save()
+
+        Group.objects.filter(name="Admin").delete()
```
janno42, line 25: why do you delete and then add the group again? you should only create it if it doesn't already exist.

T4rikA (approved):

deployment/provision_vagrant_vm.sh
```diff
@@ -13,6 +13,8 @@ echo "deb https://dl.yarnpkg.com/debian/ stable main" | sudo tee /etc/apt/source
 apt-get update
 apt-get -q install -y yarn
 
+yarn global add less
```
jeriox, line 16: @T4rikA we don't use less anymore

PR #744, opened 2020-11-07
fix wrap issue for main menu

jeriox:


janno42 (approved):
i added `mt-auto` for the management item, otherwise it would have floated in the middle

PR #754, opened 2020-11-29
Fix Empty Version Diffs

hendraet:
Apply the `migrate.sql` (e.g. by executing `\i migrate.sql` in the `dbshell`) to the database and the empty diffs should be fixed.


Paula-Kli:
Fixed with the script

PR #761, opened 2021-03-05
Update dependecies and django 3

Paula-Kli:
- channels-redis and redis was replaced with rabbitmq and channels-rabbitmq
- update also the vm to user newer python
- update to django 3
- update also other dependencies


janno42 (approved):

_1327/documents/tests.py
```diff
@@ -707,6 +707,7 @@ def setUpTestData(cls):
 		cls.document.set_all_permissions(baker.make(Group))
 
 	def test_render_text_no_permission(self):
+		print("render test no permission")
```
janno42, line 710: Only the prints need to be removed, and then we're ready to go!

Paula-Kli: done!

Issue #762, opened 2021-03-05
travis-ci.org changes to .com


Paula-Kli:
changing to github actions would be nice (maybe have a look at [EvaP](https://github.com/e-valuation/EvaP) there and copy what can be copied for that)

PR #772, opened 2021-03-14
Change travis.ci to github actions

T4rikA:
Co-authored-by: @arminWel

closes: #762 

Add tests.yml and setting_test.py for testing.
Updates requirements

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -1,3 +1,2 @@
 -r requirements-test.txt
 django-debug-toolbar >=2.0, <2.1
-psycopg2==2.8.6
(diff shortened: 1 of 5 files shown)


requirements-dev.txt
```diff
@@ -1,3 +1,2 @@
 -r requirements-test.txt
 django-debug-toolbar >=2.0, <2.1
-psycopg2==2.8.6
```
Paula-Kli, line 3: why is this removed in requirements dev?

arminWel: It is moved to  requirements-test.txt.

Paula-Kli: I see that but don't we need a Postgresql driver for development as well?

arminWel: Yep. But due to line 1, this is ensured.

Paula-Kli (approved):

